PR #10, opened 2018-08-12
CFEP-07 migration to the Anaconda compilers

mariusvniekerk:


cfep-07.md
```diff
@@ -0,0 +1,112 @@
...
+
+*   Adapt `conda-forge-pinnings` to refer to the new compilers and the new label
+
+    [PR: conda-forge-pinning-feedstock#92](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/92)
+
+    This cannot be merged to master and released until Phase 3
+
+### Phase 2
+
+*   Create a migrator using the conda-forge bot infrastructure on regro to carry out performing the
```
CJ-Wright: Should this note that we need to create the compiler subgraph? Should it also note that we will also do py3.7 and R3.5.1  and the new openBLAS, in the same migration?

CJ-Wright: Attn: @justcalamari

ocefpaf (approved):

msarahan (approved):

cfep-07.md
```diff
@@ -0,0 +1,112 @@
...
+As part of this we are introducing a few system keys into conda-forge-pinning
+
+*   channel_sources
+
+    This is a list of channels that are allowed to be used as sources for building conda-forge packages
+    
+    ```yaml
+    channel_sources:
+    - conda-forge,defaults              # [compiler_label < 7]
+    - conda-forge/label/gcc7,defaults   # [compiler_label >= 7]
```
isuruf, line 32: `conda-forge` not being here means that we have to build pure python packages, meta packages etc.

isuruf:
I guess this CFEP was accepted since https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/92 was merged.

CJ-Wright:
Maybe, I think this may still be a useful place for discussion.

minrk:
What's the plan for pure-Python / metapackages? They aren't getting rebuilt by the rerenders, but since main is excluded, they can't be found by the builds. Should non-compiled packages just get bulk-cloned over to the gcc7 channel? Or should I be triggering rebuilds of packages that don't need compilers so they get uploaded to the new channel?

mariusvniekerk:
So the plan for pure python packages is to use the upcoming strict channel priority coming in the next version of conda.
That will allow us to reasonably safely add the main label as a lower priority channel.

PR #11, opened 2018-08-13
CFEP-8: Too Big to Fail

CJ-Wright:
Proposal for how to handle packages which are too big to fail

### DIFF
--- a/cfep-08.md
+++ b/cfep-08.md
@@ -10,28 +10,28 @@
 </table>
 
 ## Abstract
+There are core packages which are "too big to fail" (TBTF). If these packages are built incorrectly they
+will cause systemic problems across the ecosystem. For example, we recently had a near-miss with the requests
+feedstock being updated to be `noarch: python` which caused [some serious issues]
+(https://github.com/conda-forge/requests-feedstock/issues/23) that, fortunately, did not have
+much of an impact *this* time.
 
-There are core packages which are "too big to fail" (TBTF). 
-These are packages, if they are built incorrectly will cause systemic problems across the ecosystem. 
-For example the recent move of some recipes to `noarch` could have caused problems with `conda` itself. 
-
-We need to have tighter control of these recipes.
-
-We need to (at least):
-1. Have all (or at least some) of core as maintainers of these packages
-1. Require review approval to merge
-
+As such, we propose that we need to have tighter control of these recipes. In practice, this 
+looks like at least:
+1. Having all (or at least some) of core as maintainers of these packages
+1. Requiring review approval to merge
 
 ## Motivation
-The move to noarch for many python projects almost crippled our conda package.
+The move to noarch for many python projects [almost crippled our conda package]
+(https://github.com/conda-forge/requests-feedstock/issues/23).
 Previous versions of `conda` would not work well with `noarch` runtime 
 dependencies. 
 This could have been particularly damaging, as a corrupted conda
 would not be able to install the needed updates to get the user out of the 
 corrupted state.
 
-In this case this was a near miss, but one could imagine changes to a 
-different part of the ecosystem doing something similar.
+In this case we had a near miss with `requests` and `conda`, but it is not difficult to 
+imagine changes to a different part of the ecosystem doing something similar.
 
 ## Implementation
 ### Controls for "Too Big to Fail" feedstocks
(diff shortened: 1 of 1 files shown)


ocefpaf:
Since we have `qt` there why not add `gdal` :wink:

CJ-Wright:
Feel free to add that under `3` (Systemic Risk). I'm sure that we all have packages which we experience as posing as a systemic risk, and I don't know if I can compile a metric to evaluate that kind of risk.

CJ-Wright:
@conda-forge/core I think this is ready for review/discussion.

ocefpaf (approved):

mwcraig (approved):
Looks good, thanks for putting it together.

cfep-08.md
```diff
@@ -0,0 +1,323 @@
...
+## Implementation
+### Controls for "Too Big to Fail" feedstocks
+Feedstocks which are TBTF would have the following additional controls on their
+updates implemented:
+1.  Have all (or at least some) of the core team as maintainers of these 
+packages
+1. Require review approval of at least one person to merge
+
+### Criteria for "Too Big to Fail"
+Any package meeting at least of the following criteria would be considered
```
bgruening: at least ... of -> is there a `one` missing?

scopatz:
Is this something that we are voting on?

CJ-Wright:
@scopatz I guess this falls under the `Overall workflow and packaging policies` clause, so yes?

scopatz:
I think we need to be more explicit about when a vote is being called for. Probably @conda-forge/core

scopatz:
Here are my thoughts on this as it stands:

1. It is certainly true that there are packages that are too big to fail and this gives some fine criteria for defining them.
2. I am not sure that I believe that "having a member of core as a maintainer" is a useful or approiate thing to do for most feedstocks, even TBTF ones.
3. I am confused as to how the other mitigation point of having a reviewer helps. Don't these PRs have to have a review already? And aren't most of the migrations PRs put out by the bot?

To elaborate on point 2 a bit more, TBTF feedstocks where there is not a core maintainer simply means that no member of core is personally invested in those packages.  I don't think that we should, by fiat, force people to be involved in a feedstock they don't care about. On the other hand, say there are maintainers (who do care) but happen to not be a members of core, and their packages suddenly cross the threshold and become TBTF. Why should a member of core be added to their maintainence team when they were doing a perfectly adequate job *before* their project breeched the TBTF criteria?

The above is basically to say that the conda-forge has a federated model, and I think this is one of its major strengths. This CFEP seems to want to give the executive (core) more control and oversight without specifiying a clear (to me) mechanism for addressing the TBTF problem. 

It seems to me that most TBTF issues are going to be wholly distinct from one another, in which case adding a core member to the mix doesn't necessarily help.

I am not trying to shoot down this CFEP as a whole. But I think it could be restructured to something less hands-on. For example:

* Core could keep a list of TBTF packages
* We could find a way to indicate that a feedstock is TBTF via the bot or adding something to the readme in rerender
* If there is a maintainer of a large number of TBTF packages, we may consider inviting them to core.

CJ-Wright:
I'll flesh this out more tomorrow, but we might consider making some tbtf recipes run their full test suite during builds and that their upstream dependencies run tests via the downstreams meta.yaml key. This might help to find issues before they become systemic and don't require additional maintainer or core overhead beyond the initial recipe fix.

cfep-08.md
```diff
@@ -0,0 +1,329 @@
...
+ 'xz',
+ 'yaml',
+ 'yasm',
+ 'zlib'}
+```
+##### 3 (Systemic Risk)
+The package has been agreed upon by core that the errors in the feedstock would 
+pose a systemic risk. These packages currently include:
+```python
+{'libgdal',
```
msarahan, line 317: I would argue that both of these are domain specific, and do not belong under "systemic risk."  They do affect a system of packages, but not the whole of conda-forge.  Should there be categories of systems?

* geospatial
* deep learning
* bio

CJ-Wright:
I think this should be tabled for the moment, so I'm going to close, but not delete the branch so we can open at another date if needed.

ericdill:
I'd change the Status to "Deferred" and merge this in.

cfep-08.md
```diff
@@ -0,0 +1,329 @@
+
+<table>
+<tr><td> Title </td><td> Packages which are "Too Big to Fail"</td>
+<tr><td> Status </td><td> Proposed </td></tr>
```
CJ-Wright:
```suggestion
<tr><td> Status </td><td> Deferred </td></tr>
```

CJ-Wright:
@ericdill ready for merge

isuruf:
Agree with @scopatz on this. For eg: I've been building the compiler packages and blas packages and nobody has been interested in helping out. That's fine by me. What's not fine is that I have to wait for somebody who is not interested in helping out when I want to fix something.

PR #12, opened 2018-08-24
CFEP 9 - Subpackage example recipe

hmaarrfk:
there seems to have been quite a bit of discussion on this topic. I think the best way to remedy this situation is to have an authoritative example on how to write a split python recipe.

I figured I would put some thoughts out there, and open it for discussion.

Pretty rough, but I think it is worth considering if this is something we want at all.


hmaarrfk:
xref: https://github.com/conda-forge/conda-forge.github.io/issues/544

PR #13, opened 2019-02-21
pinning migration cfep-9

CJ-Wright:
This is based on the awesome work by @beckermr in https://github.com/conda-forge/conda-forge.github.io/issues/712 and discusses how a new pinning system coupled with automatic pinning migrations might work.


cfep-09.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Specification
+To change the pinnings a PR will be opened into the ``migration`` directory in a proposed pinning repo.
+The PR will contain a complete copy of the global pinning file with any changes
+needed for the migration and a metadata file which includes the name of the migration,
+the source nodes which will be changed, and the finishing criteria (eg. 80% finished and 
+``numpy`` must be migrated).
+Once the commits are squashed and merged the bot will read the new file and start
+issuing PRs in topo order which bump the build number, put a copy of the modified
+pinning file into the repo for conda-smithy/build to find and rerendering the feedstock.
```
isuruf: This will override the global pinning file and then we will have different pinnings scattered in feedstocks which will not be updated using a conda-smithy rerender. (OTOH, current .ci_support./*.yaml files are updated using conda-smithy rerender)

CJ-Wright: We could have rerender clear out the file once the migration has been merged. I guess we want new pinning without a topo rebuild if the ABI hasn't broken.

CJ-Wright:
@beckermr please let me know if I failed to represent your vision properly, I'm happy to edit as needed.

cfep-09.md
CJ-Wright: It might be easier to handle a yaml snippet, since it would only override the bit of the pinning that it needs to leaving the rest to the global pinning, but then we need to use something like [``ChainDB``](https://github.com/xonsh/xonsh/blob/91438e56b09947f69bdd938a9ccf27da59f60a43/xonsh/lib/collections.py) to perform the dict merging inside of conda-build.

CJ-Wright:
@conda-forge/core I think most of the important conceptual details are here (if not implementation details) so I'd like to invite discussion on if this is a thing we want and how we might go about making this.

cfep-09.md
isuruf: How about storing these in a separate file (like `.ci_support/new_pinnings.yaml`) and make conda-smithy use the values from them only if they are newer than what is in the global pinnings file? (This is similar to what we did for py37 in conda-forge.yml file)

mariusvniekerk: So the general idea is to have a few of these that get fused at smithy time into the effective pinning file and then variant configs get generated as normal. That fusing process also deals with expiration of completed migrations etc.

mariusvniekerk: We want to have these be named somewhat arbitrariy of these as we can very likely have a number of migrations in flight simultaneously.

The other thing we want to avoid is needing to have to change both smithy and pinnings for every migration we run

isuruf: I still don't understand what changes are needed in a feedstock

beckermr:
Thanks @CJ-Wright for putting this up! 

I am reading the comments and my proposal in the thread above to make sure it all still hangs together. Some of the details on how changes to the pinnings file are ordered have been lost and these details answer some of the questions here. 

I am going to take a stab at v3 of a more detailed plan in this comment. We can translate to markdown later. This plan differs a bit from the thread above, but I think it is better. In particular, in this version, the pinnings files have a time ordered version that can be used to track state.

The migrations proceed as follows:
1. We make a PR on the migrations repo with a change to the global pinnings file, generating a new version. 
2. Once the change is merged, the bot picks it up and starts going through the graph in topo order. 
3. A migration PR is issued for a node only if 
    1. The node depends on the changed pinnings.
    2. It has no dependencies that depend on the new pinnings and have not been migrated.
    3. It is not currently being migrated by an earlier migration.
4. Process 3 continues until we determine that the migration is sufficiently complete and the change is merged into the global pinnings file. 

Some comments:
- We would need to associate a version independent of git hash with each migration that orders them in time and thus orders the pinning files in time. A pinnings file could be versioned simply with a comment inserted at the top of the YAML containing the timestamp that the change was merged or something. We could use seconds from the epoch!
- For a given migration, we only use the changes up to and including the commit that has the relevant pinning updates, even if further updates are merged for future migrations. This forces a single global pinning file to be used for a given subgraph that is being migrated. This seems required for consistency.
- We allow for a local pinnings file to be present in each feedstock that would override the global pinnings if the version is later. In other words, the more recent pinnings file is always preferred by smithy. 
- Smithy should use the most recent pinnings file when building the `.ci_support` files. This makes sure that rerendering the feedstock always produces the correct pinnings.
- The migration bot would write the new pinnings file as the local one, rerender the repo and then issue the PR.
- Migration changes have to be merged into the global pinnings file in the order they were made.

The rule of preferring the most recent pinnings file covers all of the relevant cases.
 - If a feedstock has no local pinnings, then that is fine. 
 - If the file is present, but matches the global pinnings, this is also fine.
 - In the middle of a migration, a feedstock that is being migrated would have a local pinnings file that should be used ahead of the global one. The local version would be ahead of global in this case since the migration bot would add the new pinnings file as the local one.
 - After a migration has ended and the changes from the migration are merged to the global pinnings file, then the global file will always be at the same version as the local one, or at a later version.
- If some other migration comes through later, it would then overwrite the same local pinnings file with a version that is by definition later than the current global one. 
- If a feedstock adds a dependency that has a pinned version, then the version that gets used is from the most recent pinnings file between the global one and the local one. 
- Adding a dependence in the middle of a migration would cause the repo to be detected by the bot and a migration PR would be issued.

Thus we don't need to have conda-smithy cleanup any old local pinnings files, though it could if we want.

I don't really want to allow more than one change to the pinnings of a given node at a time, mostly because we cannot guarantee that later pinning migrations can be mixed with earlier ones (e.g., both could touch the same pins for some weird reason). We need to ensure that all nodes in the subgraph that share the same pinned dependence are uniformly migrated from one version to the next. However, we can have more than one migration in flight at a time simply because they could touch different parts of the graph.

cfep-09.md
beckermr: See my comment below.

scopatz:
One thing we don't have in this disucssion (that I have seen) is a location for what migrations are curently active.  I propose that this info is put into libcflib, with a REST API.  That way conda-smithy and the auto-tick bot can query what is active at any given time.

beckermr:
Agreed @scopatz. I'd prefer to have the bot check this, but have conda-smithy only deal with pinning version numbers.

beckermr:
@CJ-Wright Can I submit a change to your PR (if you add me to your fork), or should I open a new one from my own fork with the text above?

CJ-Wright:
@beckermr I've added you as a collab to my fork.

beckermr:
OK I have moved a more detailed proposal into the markdown. Comment away!

cfep-09.md
```diff
@@ -0,0 +1,99 @@
...
+## Specification
+
+### The migrations proceed as follows:
+1. We make a PR on the `migrations` repo with a change to the global pinnings file, generating a new version. 
+2. Once the change is merged, the bot picks it up and starts going through the graph in topo order. 
+3. A migration PR is issued for a node only if 
+    1. The node depends on the changed pinnings.
+    2. It has no dependencies that depend on the new pinnings and have not been migrated.
+    3. It is not currently being migrated by an earlier migration. Per @scopatz, this state is stored in `libcflib` with a REST API.
+4. Process 3 continues until we determine that the migration is sufficiently complete and the change is merged into the global pinnings file.
```
CJ-Wright: I think we need some discussion here about how "sufficiently complete" is stated. We also might need some discussion about what to do with packages that haven't been migrated but the migration is deemed complete.

cfep-09.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Specification
+
+### The migrations proceed as follows:
+1. We make a PR on the `migrations` repo with a change to the global pinnings file, generating a new version. 
+2. Once the change is merged, the bot picks it up and starts going through the graph in topo order. 
+3. A migration PR is issued for a node only if 
+    1. The node depends on the changed pinnings.
+    2. It has no dependencies that depend on the new pinnings and have not been migrated.
+    3. It is not currently being migrated by an earlier migration. Per @scopatz, this state is stored in `libcflib` with a REST API.
```
CJ-Wright: This could cause issues of its own. One could imagine a very difficult migration (eg compiler major ABI breaks) which comes before a very simple migration. In this case we may have to wait a long time for bottleneck nodes to get through with the hard change before the simple change is implemented. This is especially painful if the simple migration is a security fix where speed is of the essence.

I don't know if the `REST` detail is needed in this point. If the bot has started migrating things, I imagine it could figure out which migrations it has open and who has claim to which nodes and in what order. The `REST` API could be important to how the bot figures out which migrations it needs to run.

cfep-09.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Rationale
+
+### Multiple migrations of the same package:
+We don't want to allow more than one change to the pinnings of a given node at a time, mostly because we cannot guarantee that later pinning migrations can be mixed with earlier ones (e.g., both could touch the same pins for some weird reason). We need to ensure that all nodes in the subgraph that share the same pinned dependence are uniformly migrated from one version to the next. However, we can have more than one migration in flight at a time simply because they could touch different parts of the graph. 
+
+### The rule of preferring the most recent pinnings file covers all of the relevant cases
+ - If a feedstock has no local pinnings, then that is fine. 
+ - If the file is present, but matches the global pinnings, this is also fine.
+ - In the middle of a migration, a feedstock that is being migrated would have a local pinnings file that should be used ahead of the global one. The local version would be ahead of global in this case since the migration bot would add the new pinnings file as the local one.
```
CJ-Wright: What if we merge a new pinning before the current migration is finished? The global pinning will be later than the local one so it will be overridden even if the newest pinning needs a topo migration.

cfep-09.md
CJ-Wright: > packages that haven't been migrated but the migration is deemed complete.

This has been an annoying thorn in the side of the compiler migration

beckermr: Right. So my feeling from watching the great compiler migration is that there is a long tail of stuff that just doesn’t get migrated. 

I think we will need settings to tell the bot how critical things are. For critical migrations we could require 100%. For minor version bumps we could be more lax.

CJ-Wright: I think we need a combination of percentage and must have packages.

tadeu:
Hi! I'd just like to point out this current issue that seems slightly related to the changes here (because it's related to the way config YAML files are handles): https://github.com/conda-forge/conda-smithy/issues/1025

Perhaps a "fix" could already come together out of the changes here?

cfep-09.md
beckermr: Yep this makes sense.

cfep-09.md
beckermr: Indeed, we could have issues around this. OTOH, if we are generating more than one migration PR per feedstock, or if successive changes in the global pinnings are incompatible, then we might have other issues. I'd say this is the least worst solution. 

We could add a mechanism to force a second migration to proceed, but we would need to develop a policy around what to do with feedstock PRs that are already open from the previous migration. We would also have to dump the time ordered versioning scheme as well to make overrides like this work.

beckermr:
Yes @tadeu. Right now the global conda-forge settings stomp on any local settings. This is probably for the best, given the purpose of global pins. However a discussion of these issues should be located in the linked issue so I will stop. ;p

cfep-09.md
beckermr: I would also add that for very intrusive/big migrations, we would want to avoid this automated system anyways in favor of work by hand. For example, if we need to do another compiler migration of every feedstock, I would argue we would want a separate set of labels for that in order to rebuild nearly everything before we release them. So I think one of the guidelines we should add here is that migrations that fall roughly under the "To Big to Fail" policy should not  be done via the global pinnings file.

cfep-09.md
beckermr: I am not fully following the question here. A few things can happen and some details. 

1. New pinnings won't be merged to the global pinnings by hand. Only after we have determined a migration is "complete", would we merge them. In this case, the strict version ordering of merging the new pinnings to the global pinnings will ensure that the correct pinnings file is used for later migrations that are in progress. In other words, by strictly ordering the changes to the global pinnings and not letting migrations jump the line in merging, we ensure that the global pinnings file and its version is always before any current migration.

2. There is the case of us merging a migration when not all feedstocks are done. In that case, the global pinnings could eventually be one or several versions ahead of the local pinnings in the unmerged migration PR. A rerender of this PR would wipe out any local pinnings. Further any maintenance on unmigrated feedstocks or any new recipes that depend on the migrated feedstocks would then get the latest pinnings and thus be migrated out of topological order.

I should add a few comments on the rationale for case 2. (I should have added these  to the markdown, sorry for not including them!)

1. I consider this a "feature", not a bug. We have to move the entire ecosystem forward at a regular cadence even if some packages fall behind. We need the least worst solution here.

2. We don't have a formal policy for deprecating feedstocks at the moment. We do archive particularly out of date feedstocks, but we leave the packages in place on anaconda.org as opposed to marking them deprecated/broken. We could handle this situation with another CFEP on regularly versioning the main label for the conda-forge channels, but that is out of scope for this CFEP. Given this, simply leaving the unmigrated packages in the main channel is consistent with what we do now and handling them properly seems out of scope.

2. This procedure is consistent with how we are currently handling the great compiler migration. Right now if someone submits a new recipe, it will be built with the new compilers even if it has a dependence that has not yet ben migrated to them. The same thing can happen to an existing feedstock if it is rerendered as part of a version bump done by hand or some other maintenance.

beckermr: The final bit here is that the assumption in all of this is that migrating a pin will produce a package that is incompatible with the unmigrated pins due to the version constraint. Thus the conda solver should do the right thing even if we force a migration to end. If a person installs an unmigrated package, then the env gets rolled back.

cfep-09.md
beckermr: I added a note on this.

cfep-09.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Implementation Details:
+- We would need to associate a version independent of git hash with each migration that orders them in time and thus orders the pinning files in time. A pinnings file could be versioned simply with a comment inserted at the top of the YAML containing the timestamp that the change was merged or something. We could use seconds from the epoch!
+- For a given migration, we only use the changes up to and including the commit that has the relevant pinning updates, even if further updates are merged for future migrations. This forces a single global pinning file to be used for a given subgraph that is being migrated. This seems required for consistency.
+- We allow for a local pinnings file to be present in each feedstock that would override the global pinnings if the version is later. In other words, the more recent pinnings file is always preferred by smithy. 
+- Smithy should use the most recent pinnings file when building the `.ci_support` files. This makes sure that rerendering the feedstock always produces the correct pinnings.
```
beckermr: One alternate path here would be to instead store the correct global pinning version for each feedstock in the conda-forge.yaml configuration file. Then upon rerender conda-smithy would need to be changed to read that version and pull the appropriate package. 

This would push all of the work for which pinning file to use to the migration bot.

beckermr: Actually no we should not do this.

cfep-09.md
CJ-Wright: > if successive changes in the global pinnings are incompatible, then we might have other issues.

I'm not certain what you mean by this. I think one of our central assumptions is that any given version of the pinning are self consistent.

CJ-Wright:
Just to make sure we're on the same page. What do we *want* to happen when we have multiple migrations at once which may touch the same package (implementation considerations aside)?

My impression is that we have three options:
1. Halt the second migration PR on that feedstock until the first is merged/closed
2. Have the second supersede the first, with the second pinning file containing the first's data
3. Merge the two migrations in a way which causes multiple builds to occur with each combination

Are their more options, which one do we want?

cfep-09.md
beckermr: I mean successive commits to the repo. We can't arbitrarily reorder git commits on the global pinnings file. Thus letting migrations hop ahead of one another might not generate the correct pinnings file.

beckermr:
I say 1. 2 is acceptable to me but could involve the bot making commits to PRs that people are actively working on. 3 is overkill since we won't need both versions in most cases.

beckermr:
Another proposal would be gut everything above and simply have the bot migrate any commit to the global pinnings file in topo order. This is a sort of "hog wild" approach, but it might not actually have that many issues associated with it. The key is to always produce consistent environments. If the pinning we adjust forces a specific version of a library to be used, then conda will take care of the rest.

scopatz:
Wait, isn't (3) the right answer above? Like what if we had a compiler migration and an OpenSSL migration? Shouldn't those both happen independently?

scopatz:
Maybe this is a fundemental question, but do pinning updates always/sometimes/never expand the build matrix?

beckermr:
Hmmmmm. So I am excluding compiler migrations from this system. Nothing above introduces a new label on anaconda.org. Instead the system I outlined above simply puts the new packages in the main channel under the assumption that the shift in pinning produces packages with different requirements for the pinned item. In this case conda will generate a correct environment. 

For migrations of critical infrastructure like compilers, I think we want separate labels and custom work.

beckermr:
Whether or not a change in pinning increases the build matrix is a matter of taste. For a simple dep, like hdf5 or something, I don't see why we need both versions. As the packages get migrated in topo order, people's envs will update themselves properly once all of the requirements are satisfied.

scopatz:
> Nothing above introduces a new label on anaconda.org.

Ahh, ok I think I misunderstood the whole point of this exercise. If it isn't adding a label then (2) seems like the right approach

beckermr:
This does raise an important question. If changes to the pinnings file don't generate packages conda can determine are incompatible, then none of the above will work. OTOH, if conda knows, then migrating in topo order is not really needed.

CJ-Wright:
Migrating in topo order will matter either way, since if we migrate the bottom first it won't get the packages it needs. Currently this causes unsatisfiable dependencies errors, making the PR error on arrival which is kinda a waste of time. The other outcome is even worse, as we might not actually rebuild the package with the rebuild libraries meaning that we think we did something and ended up doing nothing but bumping the build number.

CJ-Wright:
My main point with 2. was that if we update a pin, say openssl, then update it again right after then we shouldn't even bother with the first pin since it will be so short lived that not having packages for it might be ok. This would break things that need the first version of openssl but not the second, but that might be ok since we have collectively said the second version is the one we want to support long term.

beckermr:
Good point about topo order and dependencies being available @CJ-Wright. I'm happy to go with 2 as well.

CJ-Wright:
Based on this discussion I think the following workflow may work?
1. Propose a new pinning
2. Bot takes the new pinning file and applies it to the impacted feedstocks in topo order
3. If a new pinning is proposed send a second migration, if it impacts some of the feedstocks of the first migration then the maintainers can decide if they want to deal with both PRs or the most recent (if they just deal with the most recent then they can close the second one when the first one goes in and then that migration will also continue). (We can even try to make the bot smart about this, since it could put the ``Closes #42`` at the beginning of the newest PR which would auto close the first PR causing the first migration to go forward). This also gives the maintainers the option of doing the potentially easier first PR first, causing only the second migration to stall. 
4. Continue until criteria is met
5. Update global pinning so new packages get the new pinning.

scopatz:
Also, (2) is the "current behaviour"

beckermr:
Sounds great @CJ-Wright! I can take a pass updating the proposal later today to include all of the rationale and comments.

CJ-Wright:
Discussion from the bot meeting:
- Power users might be interested in using newer pins before a migration is finished. This could be handled by having two pining files, one which is in migration and one which is stable. Staged is built against stable, while packages can be built against either the stable (default) or migration (which is available via a cli)
- Rebuild for all pin moves vs. only ABI incompatible moves
- Have way to access the graph looking at which packages have which pins

cfep-9-migrator-file.md
```diff
@@ -0,0 +1,209 @@
...
+```
+</td>
+<td> -> </td>
+<td>
+
+```yaml
+# resulting pins
+a:
+  - 1.5
+b:
```
CJ-Wright:
```suggestion
python:
```

cfep-9-migrator-file.md
```diff
@@ -0,0 +1,209 @@
...
+  - 3.6
+  - 3.8a1
+someflag:
+  - disabled
+```
+</td>
+</tr>
+</table>
+
+These are things like the compiler migration and can be thought of as additions to a version.
```
CJ-Wright:
```suggestion
These are things like language migration and can be thought of as additions to a version.  
```

CJ-Wright: (or maybe the language portion of the compiler migration)

cfep-9-migrator-file.md
```diff
@@ -0,0 +1,209 @@
...
+```yaml
+# resulting pins
+numpy:
+  1.14
+```
+</td>
+</tr>
+</table>
+
+Key removals followed by additions are a viable way to perform a migration to a lower version.  We have never done this really.
```
CJ-Wright: My only real concern with the yaml approach is how the yaml files get bundled into migrators (eg deletion plus addition most likely should be in the same PR (ideally in the same commit))

cfep-9-migrator-file.md
```diff
@@ -0,0 +1,209 @@
...
+</td>
+<td> -> </td>
+<td>
+
+```yaml
+# resulting pins
+a:
+  - 1.6
+b:
+  - 1.2
```
beckermr: Why do versions always increase? This assumption is limiting in a way that is not required.

cfep-9-migrator-file.md
```diff
@@ -0,0 +1,209 @@
...
+  - disabled
+```
+</td>
+</tr>
+</table>
+
+These are the most common version migrations and encompass things like moving the pinnings for openssl etc.
+
+Note that the version of `b` was NOT decreased.  This is intentional as we don't wish to downgrade accidentally.  In practice many migrators could be in flight and may affect the same version.  In this case the higher version should always 
+win.  If we want to downgrade there is a mechanism but it is purposefully clunky and bad.
```
beckermr: Hmmmm. This answers my question above at some level. However I’d rather not personally be in the weeds on package versions. Changes to the global pinnings are fundamentally git file diffs.

cfep-9-migrator-file.md
```diff
@@ -0,0 +1,209 @@
...
+someflag:
+  - disabled
+```
+</td>
+</tr>
+</table>
+
+These are things like the compiler migration and can be thought of as additions to a version.  
+These need substantial care when dealing with things like zip_keys.  But in general it should be
+feasible to perform for many common cases.
```
beckermr: I’m concerned we are biting off more than we can chew here. The 80% solution of building automated tools for simpler migrations but avoiding those tools for complicated migrations seems prudent. In other words, for a complicated migration we want to have a human manually code the migration. My idea of the goal here is to have the automated migrations cover only the simplest cases where they can be effective and have a low chance of failure.

beckermr (commented):
I’ve left some specific comments above that reflect my own opinions. 

Stepping back a bit, I am a bit confused. The proposal here classifies changes to the global pinnings file into several different kinds of changes. I think the idea is that for any migration we can deduce the proper primitives and apply them.

This is all well and good but doesn’t really address how those migrations proceed. That process still involves walking the graph in topo order and applying changes to the global pinnings file. From this point of view, expressing those changes in terms of the primitives above doesn’t seem to be that helpful. The system above without the primitves can migrate an arbitrary commit. Further if we allow a node in the graph to be migrated more than once at the same time, we will have even more flexibility. Finally, the decomposition above doesn’t ensure as far as I can tell anyways that these primitves commute or can be applied in an arbitrary order. I ofc probably missed something. :(

cfep-9-migrator-file.md
mariusvniekerk: which is mostly why i'm pointing out zip_keys being a problem

cfep-9-migrator-file.md
mariusvniekerk: Mostly because for migrations decreases are almost certainly an antipattern

cfep-9-migrator-file.md
beckermr: I don't see how code that performs the operations above (with selectors which are not really adressed) really helps anything though. You are suggesting that we build and maintain code for basically what git can do for us in terms of diffs. We have to exclude complicated migrations in either case.

Did I miss something?

cfep-09.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Backwards Compatibility
+
+This proposal would require that all changes to the global pinnings file have to be routed through the new `migrations` repo. No human would make a PR on the global pinnings repo and instead that repo would be managed entirely by the bot. 
+
+## Alternatives
+
+1. Earlier versions of this proposal did not direct version the global pinnings files in a time ordered fashion. Instead they simply always preferred the local pinnings over the global ones. However, this procedure fails in the case of a packahe adding a new dependence with a pin and it having an outdated local pinnings file. By having a version stored in the file, `conda-smithy` can tell which version is more recent.
+
+2. An earlier version of this propsal was desgined around a migration being a YAML snippet. However, `conda_build_config.yaml` files can have conda-build selectors in them (e.g., `# [osx]`). This conda-build feature makes it pretty hard to use a YAML snippet as a patch on th global pinnings file. Using a direct copy of the proposed new global pinnings file is simpler. This procedure also allows us to migrate almost anything in the file.
```
jakirkham:
```suggestion
2. An earlier version of this propsal was designed around a migration being a YAML snippet. However, `conda_build_config.yaml` files can have conda-build selectors in them (e.g., `# [osx]`). This conda-build feature makes it pretty hard to use a YAML snippet as a patch on th global pinnings file. Using a direct copy of the proposed new global pinnings file is simpler. This procedure also allows us to migrate almost anything in the file. 
```

cfep-9-migrator-file.md
CJ-Wright: I think the main issue here is that git diffs require a set order to be run in. We would like to be able to apply these diffs in any order, since more than one migration could be running at a time and we don't want to block if possible.

ocefpaf:
Thanks to https://github.com/conda-forge/conda-forge.github.io/pull/786 we can call for a vote here!

@conda-forge/core please vote on CFEP-09 by approving this PR if you are :+1: or adding a :-1: in the first comment if you disapprove.

ocefpaf (approved):

CJ-Wright (commented):
I can't request changes on my own PR, but I think that the two files in these PRs need to be merged before we begin voting. I don't think the language of the two proposals are compatible.

ocefpaf:
> I can't request changes on my own PR, but I think that the two files in these PRs need to be merged before we begin voting.

Ah. I read only `cfep-09.md` :grimacing: 

> I don't think the language of the two proposals are compatible.

Is that something you are planning on doing?

beckermr:
I think we ended up blocked on the technical requirements of what we to support and not support. This needs more discussion before voting.

CJ-Wright:
@ocefpaf I can't, I'm on sabbatical while I'm working on my thesis.
I think @mariusvniekerk's approach is working so I'd go with that. I think that approach should be merged into the main `cfep-09.md` document.

beckermr:
The approach from @mariusvniekerk does not address conda selectors in the pinning files.

ocefpaf:
> @ocefpaf I can't, I'm on sabbatical while I'm working on my thesis.

Your new picture fooled me :stuck_out_tongue_closed_eyes: 
Get back to your thesis!

> I think @mariusvniekerk's approach is working so I'd go with that. I think that approach should be merged into the main cfep-09.md document.

Yep. The core meeting today was discussing @mariusvniekerk approach and it seems to be a winner.

> The approach from @mariusvniekerk does not address conda selectors in the pinning files.

OK. We need more discussion then. I'll let @mariusvniekerk answer that though.

mariusvniekerk:
@beckermr the implementation in https://github.com/conda-forge/conda-smithy/pull/1072 does support conda selectors fully

beckermr:
Ahhh great! The proposal above did not say how they would be supported which was the source of my comment. Let’s go ahead with this then.

beckermr:
This pr leaves open the issue of how to propose a new migration so that the pinning bot sees it and also how to update the global pinnings if I am not mistaken. 

We will need to work this out before merging. 

I think something close to the suggestions above will work using the migration files that will be supported when the PR above is merged.

cfep-09.md
```diff
@@ -0,0 +1,100 @@
...
+Thus we don't need to have conda-smithy cleanup any old local pinnings files, though it could if we want.
+
+
+## Backwards Compatibility
+
+This proposal would require that all changes to the global pinnings file have to be routed through the new `migrations` repo. No human would make a PR on the global pinnings repo and instead that repo would be managed entirely by the bot. 
+
+## Alternatives
+
+1. Earlier versions of this proposal did not direct version the global pinnings files in a time ordered fashion. Instead they simply always preferred the local pinnings over the global ones. However, this procedure fails in the case of a packahe adding a new dependence with a pin and it having an outdated local pinnings file. By having a version stored in the file, `conda-smithy` can tell which version is more recent.
```
pkgw: Typo: "packahe"

cfep-09.md
```diff
@@ -0,0 +1,100 @@
...
+
+2. An earlier version of this propsal was designed around a migration being a YAML snippet. However, `conda_build_config.yaml` files can have conda-build selectors in them (e.g., `# [osx]`). This conda-build feature makes it pretty hard to use a YAML snippet as a patch on th global pinnings file. Using a direct copy of the proposed new global pinnings file is simpler. This procedure also allows us to migrate almost anything in the file.
```
pkgw: Typo: "th"

cfep-09.md
```diff
@@ -0,0 +1,100 @@
...
+
+2. An earlier version of this propsal was designed around a migration being a YAML snippet. However, `conda_build_config.yaml` files can have conda-build selectors in them (e.g., `# [osx]`). This conda-build feature makes it pretty hard to use a YAML snippet as a patch on th global pinnings file. Using a direct copy of the proposed new global pinnings file is simpler. This procedure also allows us to migrate almost anything in the file. 
+
+3. Some suggestions have been centered around directly looking at the package versions in the pinnings file to determine which one is newer. This procedure has some flaws. First, it does not allow us to downgrade pinnings if needed. It is ambiguous in the case that we migrate more than one package where one version goes up and the other goes down. Finally, it does not cover changes to things like selectors or zip keys which don't have version associated with them.
```
pkgw: Typo/grammar-o: "don't have version associated with them" should probably have "versions"?

pkgw:
I know that I'm jumping into the discussion very late here, but one thing that jumped out at me on reading the proposal is that the YAML format for specifying migrations seems a little awkwardly matched to what it's trying to express. If the goal is to "decompose all of our migrations into a small set of primitives", how about having the data structure be a list of those operations?

```yaml
- type: upgrade_version
  package: mylib
  version: "1.2"
- type: downgrade_version
  package: myotherlib
  version: "1.0"
```

Maybe this is more-or-less equivalent to the current examples, but the way that the migration type is separated from the relevant package versions feels like something that's going to make it hard to express certain changes.

CJ-Wright:
@pkgw I think that part of the point here is that the actual algebra (upgrade, explosion, contraction, downgrade, etc.) is defined by the versions and the current state of the pinnings. One could imagine a situation where one package sees a downgrade where another (which hasn't been updated in ages) sees an upgrade.

CJ-Wright:
Implementations to merge:
- [x] https://github.com/conda-forge/conda-smithy/pull/1072
- [ ] https://github.com/regro/cf-scripts/pull/516
- [ ] https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/235

cfep-09.md
```diff
@@ -0,0 +1,280 @@
...
+
+2. An earlier version of this proposal was designed around a migration being a YAML snippet. However, `conda_build_config.yaml` files can have conda-build selectors in them (e.g., `# [osx]`). This conda-build feature makes it pretty hard to use a YAML snippet as a patch on th global pinnings file. Using a direct copy of the proposed new global pinnings file is simpler. This procedure also allows us to migrate almost anything in the file. 
+
+3. Some suggestions have been centered around directly looking at the package versions in the pinnings file to determine which one is newer. This procedure has some flaws. First, it does not allow us to downgrade pinnings if needed. It is ambiguous in the case that we migrate more than one package where one version goes up and the other goes down. Finally, it does not cover changes to things like selectors or zip keys which don't have a version associated with them. 
+
```
beckermr, line 293: These alternatives are outdated as the consensus solution is now using a YAML snippet.

cfep-09.md
```diff
@@ -0,0 +1,280 @@
...
+  - 1.5
+b:
+  - 1.2
+someflag:
+  - disabled
+```
+
+There are four classes of pinning we can do in general.
+
+#### 1. Simple Additive Version Migrations
```
beckermr, line 58: Does this syntax support bumping pins for only one platform? 

For example

```yaml
# migrator
__migrator: kind: version
a:
  - 1.5  # [not osx]
  - 1.6  # [osx]
b:
  - 1.1
```

CJ-Wright: @mariusvniekerk

cfep-09.md
CJ-Wright: Would it be possible to label them as such? It might be nice to document paths we considered but did not take (in case anyone asks us later about it)

beckermr: Sure. We also have the other markdown file in the PR that I think is duplicate info at this point.

cfep-09.md
CJ-Wright: I think this will work. Since the rendering (which handles the selectors) will happen first.

cfep-09.md
```diff
@@ -0,0 +1,261 @@
...
+<tr><td> Created </td><td> Feb 20, 2019</td></tr>
+<tr><td> Updated </td><td> Aug 30, 2016</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+This CFEP proposes a change to how our internal pinning system works.
+The new system will comprise of a new repo for proposed migrations.
```
beckermr: I think the abstract needs adjustment since we will use a migrations folder in the pinnings feedstock.

beckermr (commented):
Do we want an automated process to end a migration, or do we trust humans to interpret the yaml snippets correctly and then properly change the global pinnings file?

CJ-Wright:
I think that we can do this by hand for the moment. I'd be happy to start automating the process, but I think it might be best to have tighter control over the system in the beginning.

CJ-Wright:
@conda-forge/core
This PR falls under *CFEP Approval* policy, please vote and/or comment on this PR.
This PR needs *60% (10 members)* of core to vote yea to pass.
This vote will end on *Sept 6th*.

jjhelmus (approved):

isuruf:
I don't understand the downgrade strategy. For example, we have a legitimate downgrade use-case in `jpeg` where we need to downgrade to 8 to switch the underlying implementation to `libjpeg-turbo`

beckermr:
Downgrades were declared an anti pattern and as far as I know are not explicitly supported. I think maybe one could use the ordering variant for strings to force a downgrade?

beckermr:
Someone would need to consult conda smithy to test this though

CJ-Wright:
See https://github.com/conda-forge/conda-smithy/pull/1136 for downgrade test, which seems to work.

cfep-09.md
```diff
@@ -0,0 +1,303 @@
...
+  - 3.8a1
+someflag:
+  - disabled
+```
+</td>
+</tr>
+</table>
+
+These are things like the compiler migration and can be thought of as additions to a version.  
+These need substantial care when dealing with things like zip_keys.
```
isuruf, line 155: What does this mean?

beckermr: @mariusvniekerk

CJ-Wright: I think the issue here is that if you try to add a pinning under `zip_keys` then you need to make certain that the other pinnings controlled by that zip are similarly expanded otherwise they will be mismatched.

mariusvniekerk: The main concern is that it is very easy to accidentally end up with zip_keys for members that don't have the same size.    Whilst it can certainly be done using cfep9 it should be done carefully.

CJ-Wright:
@conda-forge/core for those who haven't voted yet, please do so!

cfep-09.md
beckermr: This comment has not been addressed.

CJ-Wright: Sorry about that, I think I have fix this now

scopatz (approved):

cfep-09.md
CJ-Wright: removed

mariusvniekerk (approved):

msarahan (approved):

soapy1 (approved):

bgruening (approved):

djsutherland (approved):

mwcraig (approved):

CJ-Wright:
I can't approve via github this since I'm the author, but I approve this CFEP. This gives us the 10 votes needed.

PR #21, opened 2020-02-06
Update labels in CFEP 5 to align with consensus in discussion

jasongrout:
The original CFEP 5 had the labels as `rc_package` and `dev_package`. During the [discussion of CFEP 5](https://github.com/conda-forge/cfep/pull/3), a discussion and [survey](https://github.com/conda-forge/cfep/pull/3#issuecomment-545011203) ensued about whether the prerelease indicators should be before the package name or after. The [results](https://github.com/conda-forge/cfep/pull/3#issuecomment-555629653) were that the prerelease indicators should be after, but the text of the CFEP was [never updated](https://github.com/conda-forge/cfep/pull/3#issuecomment-578368291). This commit updates the CFEP.

PR #22, opened 2020-02-06
Fix one more place in CFEP 5 where the prerelease indicator changes position

jasongrout:
This follows up on https://github.com/conda-forge/cfep/pull/21 to fix one more place that needs to change.


jasongrout:
CC @conda-forge/core

PR #23, opened 2020-02-26
CFEP-12: Removing packages that violate the terms of the source package

isuruf:


cfep-12.md
```diff
@@ -0,0 +1,33 @@
+
+<table>
+<tr><td> Title </td><td> Removing packages that violate the terms of the source package</td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Isuru Fernando &lt;isuruf@gmail.com&gt;</td></tr>
+<tr><td> Created </td><td> Feb 26, 2020</td></tr>
+<tr><td> Updated </td><td> Aug 30, 2016</td></tr>
```
CJ-Wright:
```suggestion
<tr><td> Updated </td><td> Feb 26, 2020</td></tr>
```

cfep-12.md
```diff
@@ -0,0 +1,33 @@
...
+violations are reported.
+
+## Implementation
+
+When a violation is brought to the attention of the Core team, the feedstock maintenance
+team should be made aware by the Core team in a github issue on the feedstock and notified using the
+github handle.
+
+A specified time period would be given to the maintenance team to respond to the claim and
+a specified time period would be given to address the issue.
```
pkgw: If the time periods aren't specified here, where would they be set and documented?

scopatz: Yeah, I think we should be clear here on how long we give the package

isuruf: I was thinking of them being specified on the github issue by a member of core.

scopatz: I think that creates more confusion and seems more arbitrary

beckermr (commented):
It may be useful to provide guidance on the kind of message to leave, whether or not to tag core in it, etc. to be honest, unless we have an automated way to do this I am not sure if we will be consistent in our application of the policy which is another thing to consider.

beckermr:
here is a suggestion. Let’s expand the functionality of the cf-mark-broken repo to mark all builds for an entire feedstock as broken optionally and let’s enable it to automerge certain PRs after a fixed time period, say 30 days.  

Then when we find a license issue, we can make the pr in the cf-mark-broken repo right away and leave an issue on the feedstock instructing the maintainers to fix the license and bump us on GitHub to close the mark broken pr.

scopatz (approved):

cfep-12.md
```diff
@@ -0,0 +1,33 @@
...
+violations are reported.
+
+## Implementation
+
+When a violation is brought to the attention of the Core team, the feedstock maintenance
+team should be made aware by the Core team in a github issue on the feedstock and notified using the
+github handles of the maintenance team members.
+
+A week would be given to the maintenance team to respond to the claim and
+a week would be given to address the issue.
```
beckermr: Should this say "and another week would be given to address the issue"?

isuruf: Done. Also added a sentence saying that the team may ask for it to be extended.

beckermr:
Any thoughts on automation here @isuruf?

beckermr:
Or did we hit this again: https://xkcd.com/1205/

isuruf:
I don't think we need to talk about automation here, but we can iterate on a draft message here if you'd like to.

cfep-12.md
```diff
@@ -0,0 +1,34 @@
...
+github handles of the maintenance team members.
+
+A week would be given to the maintenance team to respond to the claim and
+another week would be given to address the issue. The feedstock maintenance team
+may ask to extend the deadlines.
+
+If the issue is not addressed after the specified time period, the package will be moved
+to the broken label.
+
+
```
beckermr, line 32:
```suggestion
We suggest a message to the maintainers along the following lines

    Hi! We (`@conda-forge/core`) have noticed that this package does not have
    the license file required for us to be able to distribute this code. Nearly every
    open-source license has such a requirement. It is extremely important that we
    at `conda-forge` respect license restrictions and requirements such as this one
    so that we can be good stewards of the open-source community. 
    
    We are requesting that you add the correct license file to this feedstock. 
    If this is not done within two weeks, the packages from this feedstock be 
    moved to the broken label. If you'd like to extend this deadline or need 
    to discuss something with us, please feel free to respond to this issue.
    
    We also very much appreciate the time you have taken to maintain this 
    feedstock and your understanding on this issue!
```

beckermr: @isuruf

isuruf:
@conda-forge/core

This vote will end on 2020-03-11.

beckermr (approved):

sodre (dismissed):

pkgw (approved):

mwcraig (approved):

synapticarbors (approved):

cfep-12.md
```diff
@@ -0,0 +1,48 @@
...
+When a violation is brought to the attention of the Core team, the feedstock maintenance
+team should be made aware by the Core team in a github issue on the feedstock and notified using the
+github handles of the maintenance team members.
+
+A week would be given to the maintenance team to respond to the claim and
+another week would be given to address the issue. The feedstock maintenance team
+may ask to extend the deadlines.
+
+If the issue is not addressed after the specified time period, the package will be moved
+to the broken label.
```
ericdill: If I'm understanding correctly, then I don't think that moving the package to the broken label truly addresses the redistribution violation. We need to **remove** all binaries from the conda-forge channel until the redistribution violation is addressed. Otherwise we, conda-forge, are still redistributing the conda package that is in violation of the source packages terms and conditions. The "broken" label doesn't actually protect conda-forge from a redistribution violation.

If I'm understanding correctly, then 41 needs to be updated to say that we're going to _remove_ the package from conda-forge.

ericdill (changes requested):
Thanks for codifying this @isuruf 

tl;dr. I'd like to sort out this main issue that I see: moving packages to `broken` doesn't fix redistribution violations. If we're concerned about these violations, we should delete the package from anaconda.org/conda-forge until it's fixed.

msarahan:
What about PR's from the bot?  It seems like any feedstock deemed out of compliance shouldn't be allowed to have new packages until they have proven that they have addressed their license issue.  As it stands, their *current* packages are moved to broken, but what about any updates built afterwards?

jakirkham:
I thought the bot already had something in place to add the license file to recipes that don't have it.

It seems to me if we are really concerned about this we could just have the bot add license files to every feedstock missing them.

beckermr:
Deleting all builds and then archiving the feedstock should address both of these concerns. I’d want a longer grace period for more extreme measures like this maybe.

beckermr:
A list of feedstocks sans license files is located in the bot data. I forget where.

jakirkham:
Right we have a piggyback migrator for `license_file`s already though. PR ( https://github.com/regro/cf-scripts/pull/637 ) introduced if I'm not mistaken. Would it make sense to just make it a normal migrator? Are there fail cases in it that we can identify?

isuruf:
@jakirkham, I don't see how that is relevant. Some feedstocks might have inadequate `license_file` entries. (For eg: a package's license is in 2 files and we only package 1.) Some packages might not be redistributable at all. @msarahan was suggesting we stop the bot from making version update PRs, etc if I understood him correctly.

jakirkham:
I'm not sure I follow. The issue is packages not having license files, right?

We can either be punitive and remove packages without license files or we can be constructive and make sure we add them across the stack. Leveraging the bot let's us constructively solve the problem. That is what I'm suggesting.

So we never fix these feedstocks? I'm not sure I agree with that proposal.

isuruf:
> The issue is packages not having license files, right?

Not entirely. Please see my comment https://github.com/conda-forge/cfep/pull/23#issuecomment-594378231 where I describe two scenarios where the existing license_file is inadequate.

jakirkham:
So what I’m hearing is you would like the license migrator to handle multiple license files 😉

isuruf:
Not everything can be automated easily. For eg: https://github.com/conda-forge/pycosat-feedstock/issues/13 is missing a license text from the header. How can a bot easily identify this?

cfep-12.md
ocefpaf: Yep. broken label is still a channel and the package is still available. This is probably the only case were a removal is the right procedure.

ocefpaf (changes requested):
The package most be removed and not moved to the broken label.

With that said I thought we agreed on waiting for upstream to send take down notices instead of actively find and remove packages with bad licenses.

cfep-12.md
jjhelmus: From a legal stand point I agree with @ericdill, deleting the package is the only way to stop re-distribution. 

From a user's stand point moving to `broken` is more attractive as it still allows access to the package when re-creating a historic environment.

Perhaps a compromise is to move the package to `broken` by default but remove the package if the upstream developers request this action.  My though is that many developers would be happy to have the license violation corrected even if packages exist which have a violation and are marked as such.

Do other packaging communities (Debian, CentOS, Arch) remove packages from their repositories when a license violation is found or do they release new builds to address the issue?

xhochy: Moving things "only to broken" could also get us into more trouble as we show that we know that we are aware of an issue but didn't take appropriate action.

ocefpaf: Both @jjhelmus and @xhochy points goes back to what I asked above. Shouldn't we take action if, and only if, we get a take down notice? We can ask NumFOCUS about it.

beckermr: I suggested we consult a lawyer on gitter and still think we should do that.

ocefpaf: I recall that the decision to wait on a take down notice was taken after we spoke with NumFOCUS people and Anaconda. We can still ask NumFOCUS about. I don't remember why we ended up not asking them to forward that to the lawyer.

PS: who is out primary NumFOCUS contact these days?

jjhelmus: Debian has some documentation on their removal process, https://www.debian.org/doc/manuals/developers-reference/pkgs.en.html#removing-pkgs.  

From my reading package are deleted from un-released distributions, ( `unstable`, `experimental` and `stable`) but not from released distributions (`jessie`, `stretch`, `buster`).

Not sure how or if this should translate to conda-forge which has little idea of a "released distribution".  Perhaps if a package is included in miniforge it should not be removed?

ocefpaf: @jjhelmus that does not say anything about removal due license violations. (Or I missed while reading it.)

If I recall correctly Debian has a very pick way of including packages and license issues are solved before they make into a distribution.

jjhelmus (approved):

mbargull (changes requested):
It pains me do this, but I'm rejecting this proposal for now.
I want us to improve on (well, ideally fix, actually) all the licensing issues there are on conda-forge sooner than later.
But I also want us to "get it right", e.g.,
- do it with as less end-user impact (e.g., for reproducibility) as possible,
- not stepping into "traps" like mentioned in https://github.com/conda-forge/cfep/pull/23#discussion_r387755549,
- have clarification on what the `license` field is supposed to mean (source/package license, ref: https://github.com/conda-forge/sqlite-feedstock/pull/38#discussion_r218827797)
- etc.

I am thankful for everyone that looks into licensing issues, which very much includes @isuruf who not only put up this proposal but I know is also on the lookout for those on our feedstocks.

However, I'd like us to take some more time on deciding on this because it *is such an important topic*.

isuruf:
How about this?

1. Make the package private (only core can see) if the deadline passes.
2. Hotfix the older packages to include a `<pkg>-license` which is generated by a script if the recipe is fixed.

ocefpaf:
> 1. Move the packages to a private channel if the deadline passes

I'm not sure we can do that. It half solves the reproducibility issue and may still violates some licenses (IANAL BTW).

> 2. Hotfix the older packages to include a `<pkg>-license` which is generated by a script if the recipe is fixed.

That should definitely be in here!

jakirkham:
> Not everything can be automated easily. For eg: [conda-forge/pycosat-feedstock#13](https://github.com/conda-forge/pycosat-feedstock/issues/13) is missing a license text from the header. How can a bot easily identify this?

It's a good question and likely with an evolving answer. 🙂

Mostly I think people poke around on GitHub or download and extract source archives to find something that says `LICENSE*`, `COPY*`, etc. This behavior seems automatable, but you are right we will likely discover edge cases. 🤔

FWIW it seems grayskull is trying to solve this problem ( https://github.com/marcelotrevisani/grayskull/pull/58 ). Maybe this would be a good place for further engagement? 😉

msarahan (changes requested):
@mbargull said it perfectly

jjhelmus:
I'd like to see a resolution to the question of moving to `broken` vs deleting the package and/or a proposed that recommends a workflow based on @mbargull and @isuruf comments before I approve this CFEP.

jjhelmus (changes requested):
I'd like to see a resolution to the question of moving to broken vs deleting the package and/or a proposed that recommends a workflow based on @mbargull and @isuruf comments before I approve this CFEP.

isuruf:
I've addressed all of the concerns except for the following.

> have clarification on what the license field is supposed to mean

That's unrelated to this CFEP.

jakirkham (changes requested):
This still seems to have issues with environment reproducibility. I'm not sure how we fix that without continuing to have the packages available to external people at some level. So for now am voting against this proposal. Open to hearing solutions to this problem though

ocefpaf:
> This still seems to have issues with environment reproducibility. I'm not sure how we fix that without continuing to have the packages available to external people at some level. So for now am voting against this proposal. Open to hearing solutions to this problem though

@jakirkham the first step is to ask NumFOCUS for legal advice here. Once we know what we can and cannot do then it will be easier to lay a path forward. I'm looking into that.

isuruf:
> I'm not sure how we fix that without continuing to have the packages available to external people at some level.

This feels like saying, hey I did something illegal, but please let me keep doing it because I did it in the past.

ocefpaf:
> This feels like saying, hey I did something illegal, but please let me keep doing it because I did it in the past.

Yep. Like we discussed in the last meeting. If we are going to do this right we'll have to make peace with breaking envs. That is the whole reason why I want to touch base with a lawyer first.

isuruf:
FYI, my proposal here doesn't break envs unless there's a serious issue or if the maintainers are unwilling to fix it. If there's a serious issue about packaging, we can't do anything about it. (There's the check sum issue, but that's minor and people can update those.)

If the maintainers are unwilling to fix it, then people who are using it will notice that their environments broke and they might help with fixing it.

jjhelmus:
Changing the contents of packages is a bad idea in my opinion, I would much prefer a new build.  Changing the contents of a package requires the original package to be invalidated in a number of places including anaconda.org, the CDN and on user's local package cache.  Each of these can be tricky and error prone.

isuruf:
I checked anaconda.org and we can actually do that and the CDN uses anaconda.org I believe. I don't know what goes behind anaconda.org, so I don't know how that's going to affect anaconda.org, though.
User local package cache is fine as we are not redistributing it.

msarahan:
It would be creating a lot of work for the distribution team.  It also has been disruptive to users behind caching proxies in the past.  These people rarely even know that they're behind a caching proxy - things just start giving them weird checksum mismatch errors.

I don't think there's any way to have both licenses added to old packages and reproducibility.  One possible workaround would be to allow sidecar files that could augment the package metadata.  It could have hotfix stuff as well as license additions/changes.

Edit: historical context: https://github.com/conda/conda/issues/4956

isuruf:
>  One possible workaround would be to allow sidecar files that could augment the package metadata. It could have hotfix stuff as well as license additions/changes.

This wouldn't fix the comments made above that we are still distributing the conda package without these files in anaconda.org.

msarahan:
Yes, you can't have your cake and eat it, too.

jjhelmus:
Conda will give a `MD5MismatchError` when the contents of the local cache do not match those in repodata.json.  Every time packages contents were changed in `defaults` we had users report errors along these lines. It is possible to remove the index cache using `conda clean -iy` but requiring that is a poor user experience.

jjhelmus:
> Edit: historical context: [conda/conda#4956](https://github.com/conda/conda/issues/4956)

Thanks for finding that issue @msarahan.  I was searching for it but could not find it.

beckermr:
We really need good legal advice here. Let's hold off on further discussion until we hear back. Then I propose we dedicate time on the phone to go through the advice from the lawyers on this.

CJ-Wright:
What is the impact of this policy in terms of number of feedstocks and packages that need fixing, number of child packages that couldn't be built and number of downloads that's would be impacted?

jakirkham:
We can probably only get an approximate number (like what things need `license_file`). Though it still doesn't cover some edge cases that @isuruf mentioned. It also would include things that don't technically need a license file (like Public Domain).

isuruf:
`license_file` is not related to this CFEP. packages that don't have a license_file doesn't necessarily mean that the conda package violates the terms. Some packages distribute the license in $PREFIX/share. Some packages are source only and have the copyright and license as a header of those source files.
So, let's not drag `license_file` into this CFEP.

beckermr:
I see the whole holding on further discussion thing is not going to fly. ;)

isuruf:
It has been a month now. Did we hear back from NumFOCUS?

isuruf:
Another 1.5 months have passed now.

ocefpaf:
> Another 1.5 months have passed now.

I bet it will be longer. These things are slow and we must be patient. Last week I got a feedback that our second communication to NumFOCUS about this was forwarded to the legal department. They are evaluating now and will probably come back with questions, not answers.

cfep-12.md
```diff
@@ -0,0 +1,87 @@
...
+      have to be deleted or made private according to the previous guidelines.
+
+
+We suggest a message to the maintainers along the following lines
+
+    Hi! We (`@conda-forge/core`) have noticed that this package does not have
+    the license file required for us to be able to distribute this code. Nearly every
+    open-source license has such a requirement. It is extremely important that we
+    at `conda-forge` respect license restrictions and requirements such as this one
+    so that we can be good stewards of the open-source community.
```
moorepants, line 76:
```suggestion
    so that we can be good stewards of the open-source community and follow the law.
```

PR #24, opened 2020-02-28
[CFEP-13] secure package uploads

beckermr:
This PR has the secure package uploads procedure we discussed. 

cc @conda-forge/core for comments

### DIFF
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -44,10 +44,13 @@ We will secure the uploads of outputs to `anaconda.org` via the following proces
 In order to mitigate the impact on feedstock maintainers, we will
 
 1. Advertise this change widely as it is implemented.
-2. Create an admin bot command to create a PR for new outputs.
+2. Create an admin bot command to create a PR for new registering new outputs.
 3. Add a linter check that all outputs are registered for this feedstock in the global database.
 4. Automerge PRs for outputs that don't overlap with any
    existing name on any feedstock.
+   
+The above combination of items will let users register new outputs for feedstocks 
+in a mostly self-service manner while also not letting them write to any output.
 
 ## Alternatives
 
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -47,8 +47,8 @@ In order to mitigate the impact on feedstock maintainers, we will
 1. Advertise this change widely as it is implemented.
 2. PR CI runs should check all of their outputs against the web service in a "dry-run"
    mode and fail if any output conflicts with some other package.
-3. Both the "dry run" and "upload" commands to the web service will automatically 
-   register new outputs that do not conflict.
+3. Package "upload" commands to the web service will automatically 
+   register new outputs for a feedstock that do not conflict with any other outputs.
    
 The above combination of items will let users register new outputs for feedstocks 
 in a mostly self-service manner while also not letting them write to an arbitrary 
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -45,7 +45,7 @@ In order to mitigate the impact on feedstock maintainers, we will
 
 1. Advertise this change widely as it is implemented.
 2. Create an admin bot command to create a PR for new outputs.
-3. Add a linter check that all outputs exist.
+3. Add a linter check that all outputs are registered for this feedstock in the global database.
 4. Automerge PRs for outputs that don't overlap with any
    existing name on any feedstock.
 
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -34,9 +34,10 @@ We will secure the uploads of outputs to `anaconda.org` via the following proces
    outputs, their SHA256 checksums, the feedstock name, and the secret key to a
    validation web service. This web service then verifies the secret key-feedstock
    combination and verifies that the outputs are allowed for that feedstock.
-4. If everything is ok, the outputs are copied from the staging organization to the
+5. If everything is ok, the outputs are copied from the staging organization to the
    main conda-forge channels.
-6. The service should post a comment on the PR that the outputs were copied.
+6. The service should fail the CI run if the output is not copied. It should 
+   also post a comment on the PR that the outputs were copied or not to help users.
 
 
 ### The Impact on Users
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -1,10 +1,10 @@
 
 <table>
 <tr><td> Title </td><td> Securing `conda-forge` Uploads to `anaconda.org` </td>
-<tr><td> Status </td><td> Proposed </td></tr>
+<tr><td> Status </td><td> Accepted </td></tr>
 <tr><td> Author(s) </td><td> Matt, Marius, Anthony, Isuru</td></tr>
 <tr><td> Created </td><td> Feb 27, 2020</td></tr>
-<tr><td> Updated </td><td> Feb 27, 2020</td></tr>
+<tr><td> Updated </td><td> Mar 12, 2020</td></tr>
 <tr><td> Discussion </td><td> NA </td></tr>
 <tr><td> Implementation </td><td> NA </td></tr>
 </table>
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -23,20 +23,19 @@ We will secure the uploads of outputs to `anaconda.org` via the following proces
 
 1. Each feedstock is provisioned with a secret key that is only known to the
    maintainers of that feedstock and the core team. These keys will be kept in
-   a private repo with only core members having read and write access. This key will be pushed to
-   the CI providers for that repo as well.
-2. Feedstock creation will move to a separate process in order to secure the keys
-   associated with our CI provider accounts and to securely register outputs of feedstocks.
-3. A public, global registry of all outputs from all feedstocks will be kept
+   a private repo with only core members having read access. Core members and 
+   encrypted GitHub keys for bot admins will have write access in order to generate 
+   the secret keys and push then to the CI providers.
+2. A public, global registry of all outputs from all feedstocks will be kept
    in a repo on github. Write access to this repo is limited to core as well.
-4. When a feedstock builds new outputs it wants to upload, it will push these outputs
+3. When a feedstock builds new outputs it wants to upload, it will push these outputs
    to a staging organization on `anaconda.org`. It will also send the names of the
    outputs, their SHA256 checksums, the feedstock name, and the secret key to a
    validation web service. This web service then verifies the secret key-feedstock
    combination and verifies that the outputs are allowed for that feedstock.
-5. If everything is ok, the outputs are copied from the staging organization to the
+4. If everything is ok, the outputs are copied from the staging organization to the
    main conda-forge channels.
-6. The service should fail the CI run if the output is not copied. It should 
+5. The service should fail the CI run if the output is not copied. It should 
    also post a comment on the PR that the outputs were copied or not to help users.
 
 
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -27,7 +27,8 @@ We will secure the uploads of outputs to `anaconda.org` via the following proces
    encrypted GitHub keys for bot admins will have write access in order to generate 
    the secret keys and push then to the CI providers.
 2. A public, global registry of all outputs from all feedstocks will be kept
-   in a repo on github. Write access to this repo is limited to core as well.
+   in a repo on github. Write access to this repo is limited to core and bot 
+   admins as well.
 3. When a feedstock builds new outputs it wants to upload, it will push these outputs
    to a staging organization on `anaconda.org`. It will also send the names of the
    outputs, their SHA256 checksums, the feedstock name, and the secret key to a
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -27,10 +27,10 @@ We will secure the uploads of outputs to `anaconda.org` via the following proces
 2. When a feedstock builds new outputs it wants to upload, it will push these outputs
    to a staging organization on `anaconda.org`. These outputs will either be signed 
    so that they can be validated or the feedstock will directly alert the validation 
-   service.
+   service in a secure, verifiable manner.
 3. The validation service will validate that any outputs in the staging org are 
    from the correct feedstock, the actual outputs to be upoloaded, and allowed 
-   for the feedstock by checking against the global registry.
+   for the feedstock (given the global registry).
 4. If everything is ok, the outputs are copied from the staging organization to the
    main conda-forge channels. The outputs are then deleted once the copy is done.
 5. The service should report a failure to users if an output is not copied.
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -21,7 +21,8 @@ should mitigate.
 
 We will secure the uploads of outputs to `anaconda.org` via the following process.
 
-1. A public, global registry of all outputs from all feedstocks will be kept
+1. A public, global registry of the package names mapped to the feedstock that 
+   that generated them will be kept
    in a repo on github. Write access to this repo is limited to core and bot 
    admins as well.
 2. When a feedstock builds new outputs it wants to upload, it will push these outputs
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -29,7 +29,7 @@ We will secure the uploads of outputs to `anaconda.org` via the following proces
    associated with our CI provider accounts and to securely register outputs of feedstocks.
 3. A public, global registry of all outputs from all feedstocks will be kept
    in a repo on github. Write access to this repo is limited to core as well.
-3. When a feedstock builds new outputs it wants to upload, it will push these outputs
+4. When a feedstock builds new outputs it wants to upload, it will push these outputs
    to a staging organization on `anaconda.org`. It will also send the names of the
    outputs, their SHA256 checksums, the feedstock name, and the secret key to a
    validation web service. This web service then verifies the secret key-feedstock
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -23,7 +23,7 @@ We will secure the uploads of outputs to `anaconda.org` via the following proces
 
 1. Each feedstock is provisioned with a secret key that is only known to the
    maintainers of that feedstock and the core team. These keys will be kept in
-   a private repo with only core members having write access. This key will be pushed to
+   a private repo with only core members having read and write access. This key will be pushed to
    the CI providers for that repo as well.
 2. Feedstock creation will move to a separate process in order to secure the keys
    associated with our CI provider accounts and to securely register outputs of feedstocks.
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -35,7 +35,7 @@ We will secure the uploads of outputs to `anaconda.org` via the following proces
    validation web service. This web service then verifies the secret key-feedstock
    combination and verifies that the outputs are allowed for that feedstock.
 4. If everything is ok, the outputs are copied from the staging organization to the
-   main conda-forge channels.
+   main conda-forge channels. The outputs are then deleted once the copy is done.
 5. The service should fail the CI run if the output is not copied. It should 
    also post a comment on the PR that the outputs were copied or not to help users.
 
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -45,14 +45,14 @@ We will secure the uploads of outputs to `anaconda.org` via the following proces
 In order to mitigate the impact on feedstock maintainers, we will
 
 1. Advertise this change widely as it is implemented.
-2. Create an admin bot command to create a PR for new registering new outputs.
-3. PR CI runs should check all of their outputs against the web service in a "dry-run"
-   mode and fail if any output is not registered.
-4. PRs for registering new outputs will be automatically merged if they 
-   don't overlap with any existing name on any feedstock.
+2. PR CI runs should check all of their outputs against the web service in a "dry-run"
+   mode and fail if any output conflicts with some other package.
+3. Both the "dry run" and "upload" commands to the web service will automatically 
+   register new outputs that do not conflict.
    
 The above combination of items will let users register new outputs for feedstocks 
-in a mostly self-service manner while also not letting them write to any output.
+in a mostly self-service manner while also not letting them write to an arbitrary 
+output.
 
 ## Alternatives
 
--- a/cfep-13.md
+++ b/cfep-13.md
@@ -29,7 +29,7 @@ We will secure the uploads of outputs to `anaconda.org` via the following proces
    so that they can be validated or the feedstock will directly alert the validation 
    service in a secure, verifiable manner.
 3. The validation service will validate that any outputs in the staging org are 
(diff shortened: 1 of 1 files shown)


cfep-13.md
```diff
@@ -0,0 +1,64 @@
...
+3. A public, global registry of all outputs from all feedstocks will be kept
+   in a repo on github. Write access to this repo is limited to core as well.
+3. When a feedstock builds new outputs it wants to upload, it will push these outputs
+   to a staging organization on `anaconda.org`. It will also send the names of the
+   outputs, their SHA256 checksums, the feedstock name, and the secret key to a
+   validation web service. This web service then verifies the secret key-feedstock
+   combination and verifies that the outputs are allowed for that feedstock.
+4. If everything is ok, the outputs are copied from the staging organization to the
+   main conda-forge channels.
+6. The service should post a comment on the PR that the outputs were copied.
```
isuruf: How about just waiting for the webservice to report that the package was copied?

cfep-13.md
```diff
@@ -0,0 +1,64 @@
...
+
+1. Each feedstock is provisioned with a secret key that is only known to the
+   maintainers of that feedstock and the core team. These keys will be kept in
+   a private repo with only core members having write access. This key will be pushed to
+   the CI providers for that repo as well.
+2. Feedstock creation will move to a separate process in order to secure the keys
+   associated with our CI provider accounts and to securely register outputs of feedstocks.
+3. A public, global registry of all outputs from all feedstocks will be kept
+   in a repo on github. Write access to this repo is limited to core as well.
+3. When a feedstock builds new outputs it wants to upload, it will push these outputs
```
isuruf:
```suggestion
4. When a feedstock builds new outputs it wants to upload, it will push these outputs
```

cfep-13.md
```diff
@@ -0,0 +1,64 @@
...
+   main conda-forge channels.
+6. The service should post a comment on the PR that the outputs were copied.
+
+
+### The Impact on Users
+
+In order to mitigate the impact on feedstock maintainers, we will
+
+1. Advertise this change widely as it is implemented.
+2. Create an admin bot command to create a PR for new outputs.
```
isuruf: Do we need this? Maybe the webservice should just create a PR for new outputs.

cfep-13.md
```diff
@@ -0,0 +1,64 @@
...
+3. Add a linter check that all outputs exist.
```
isuruf: What does this mean?

cfep-13.md
beckermr: Not sure what you mean. Once the web service has copied, it should report back to let people know.

cfep-13.md
beckermr: Well for a package to be uploaded, all of its outputs must be in the global database. The linter should check this and warn users if not.

cfep-13.md
isuruf: In CI, we send a request to the webservice and the webservice can send a 200 code if the package was copied and a 403 if it's forbidden. CI could fail if the code is not 200.

cfep-13.md
beckermr: Well in principle, the outputs need to be in the database before the webservice attempts to copy them. So the idea here is that if a user gets a linter warning that an output is not registered, they can use the admin command to generate the PR to do so. This PR will get merged automatically if nothing conflicts.

cfep-13.md
isuruf:
```suggestion
3. Add a linter check that all outputs are registered for this feedstock in the global database.
```

cfep-13.md
beckermr: Ah I see. That is fine for the CI, but it doesn't really help people understand why something failed. I'd like things to be transparent enough so that users know which thing to go track down when we inevitably don't copy something due to a bug, etc.

cfep-13.md
```diff
@@ -0,0 +1,64 @@
...
+6. The service should post a comment on the PR that the outputs were copied.
+
+
+### The Impact on Users
+
+In order to mitigate the impact on feedstock maintainers, we will
+
+1. Advertise this change widely as it is implemented.
+2. Create an admin bot command to create a PR for new outputs.
+3. Add a linter check that all outputs are registered for this feedstock in the global database.
```
isuruf: This would make the linter as slow at the rerender command because the package has to be rendered for all the platforms.

beckermr: Hmmmm. We could allow the CI on PRs to do a test check of the outputs against the web service and the CI can fail for PRs if outputs are not registered.

beckermr: My main thing is that something somewhere needs to report in a visible way to users what they need to do.

beckermr: We could have the rerender command do this check and post a comment. That is slow anyways and actually renders the recipe.

isuruf: +1. I would remove the linter check then add something about using CI for this.

isuruf: > We could have the rerender command do this check and post a comment. That is slow anyways and actually renders the recipe.

You mean run rerender on every push to a PR? That would cripple the webservice

beckermr: No I mean when it happens to be called. So if a user happens to use the rerender service to rerender a feedstock, we go ahead and check the outputs then.

beckermr: The CI should check too.

isuruf: That's too much complexity. Just CI is enough IMO.

beckermr: fair enough - I guess it can post a loud error message that people can look at

cfep-13.md
```diff
@@ -0,0 +1,69 @@
...
+### The Impact on Users
+
+In order to mitigate the impact on feedstock maintainers, we will
+
+1. Advertise this change widely as it is implemented.
+2. Create an admin bot command to create a PR for new registering new outputs.
+3. PR CI runs should check all of their outputs against the web service in a "dry-run"
+   mode and fail if any output is not registered.
+4. PRs for registering new outputs will be automatically merged if they 
+   don't overlap with any existing name on any feedstock.
```
isuruf: Do we need PRs in this case? Just push to master?

cfep-13.md
```diff
@@ -0,0 +1,69 @@
...
+
+
+### The Impact on Users
+
+In order to mitigate the impact on feedstock maintainers, we will
+
+1. Advertise this change widely as it is implemented.
+2. Create an admin bot command to create a PR for new registering new outputs.
+3. PR CI runs should check all of their outputs against the web service in a "dry-run"
+   mode and fail if any output is not registered.
```
isuruf: If there's no overlap, we shouldn't fail

beckermr: Well something somewhere has to push to the registry of outputs and this has to be gated in some way.

cfep-13.md
beckermr: Yes we can have the admin command push straight to master if there is no overlap.

cfep-13.md
beckermr: The failure is here to indicate to a user that they need to do something to register their outputs.

isuruf: The webservice that validates the outputs can push to the registry if there's no overlap. This way, only users who add outputs that conflict needs to do anything. Others can be oblivious of this change.

beckermr: I guess given everything else the dry run mode could simply register outputs that don't exist and only fail if there is a conflict.

beckermr: this is no different than automerge

beckermr: ditto for uploads

isuruf: > this is no different than automerge

Exactly. I wouldn't register in a PR though as people make mistakes.

beckermr: Ack sorry I missed your comments @isuruf! Yup I agree!

cfep-13.md
```diff
@@ -0,0 +1,69 @@
...
+
+
+### The Impact on Users
+
+In order to mitigate the impact on feedstock maintainers, we will
+
+1. Advertise this change widely as it is implemented.
+2. PR CI runs should check all of their outputs against the web service in a "dry-run"
+   mode and fail if any output conflicts with some other package.
+3. Both the "dry run" and "upload" commands to the web service will automatically
```
isuruf: I'm not a fan of registering in `dry run` mode as people make mistakes in the PRs or change their minds about naming.

beckermr: good point

cfep-13.md
```diff
@@ -0,0 +1,69 @@
...
+
+### The Impact on Users
+
+In order to mitigate the impact on feedstock maintainers, we will
+
+1. Advertise this change widely as it is implemented.
+2. PR CI runs should check all of their outputs against the web service in a "dry-run"
+   mode and fail if any output conflicts with some other package.
+3. Package "upload" commands to the web service will automatically 
+   register new outputs for a feedstock that do not conflict with any other outputs.
```
isuruf:
```suggestion
   register new outputs for a feedstock that do not conflict with any other feedstock.
```

beckermr: Well it is the outputs of any other feedstock.

beckermr: take a look at the new text which combines both things and is most clear I hope

cfep-13.md
```diff
@@ -0,0 +1,70 @@
...
+of say the `python` package. This situation is a security risk that we
+should mitigate.
+
+## Specification
+
+We will secure the uploads of outputs to `anaconda.org` via the following process.
+
+1. Each feedstock is provisioned with a secret key that is only known to the
+   maintainers of that feedstock and the core team. These keys will be kept in
+   a private repo with only core members having write access. This key will be pushed to
```
isuruf: I don't think we need maintainers any access at all. They'd have indirect access via CI, but that's it.

```suggestion
   a private repo with only core members having read and write access. This key will be pushed to
```

cfep-13.md
```diff
@@ -0,0 +1,70 @@
...
+   the CI providers for that repo as well.
+2. Feedstock creation will move to a separate process in order to secure the keys
```
isuruf: What does this mean? Why can't we have the same process as now?

cfep-13.md
beckermr: Yes this was my intent with the phrase "private repo" but it was not clear. 100% agreed!

cfep-13.md
beckermr: Well @mariusvniekerk seemed to suggest on gitter that this would be a good change, but I agree the exact technical issue here is not clear to me. 

We do need to generate and register new keys to the private repo in a secure way. IDK if having this done with code on staged recipes is a great idea. 

However, we don't have to redo staged recipes to address this. We could have the web service watch repo creation events in the conda-forge org and automatically register a key when it detects a new feedstock.

beckermr: We also need to add the keys to feedstock CI services in a secure way. Again it is not 100% clear to me how we do this. Maybe the webservice has a register command we call.

isuruf: Why can't we have the feedstock creation script in `staged-recipes` do this?

beckermr: The actual constraint is that we have to avoid leaking the secret key we generate to the outside world. 

The question of who can see those keys is separate at some level. Having them be generated in staged recipes would allow core+any staged recipe maintainer be able to see the keys as they are made in principle.

If we are ok with that, then staged-recipes is fine. 

After thinking about it for a while and remembering that I was not always on core, I think I'd still prefer to not have even our encrypted tokens for any service available outside of core. However, I agree this is a different issue that is orthogonal to this CFEP.

isuruf: I agree that this is orthogonal to this CFEP. (If we go this route, we'd need to remove the feedstock creation scripts from staged-recipes as they contain github tokens, CI tokens, etc.)

cfep-13.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Specification
+
+We will secure the uploads of outputs to `anaconda.org` via the following process.
+
+1. Each feedstock is provisioned with a secret key that is only known to the
+   maintainers of that feedstock and the core team. These keys will be kept in
+   a private repo with only core members having read access. Core members and 
+   encrypted GitHub keys for bot admins will have write access in order to generate 
+   the secret keys and push then to the CI providers.
```
scopatz: Shouldn't this be in keybase rather than on github?

beckermr: We can do this, but IDK how to push and pull from keybase with a key securely when that key is stored on staged-recipes.

beckermr: We may not want keys for keybase itself floating around in our CI scripts.

isuruf (approved):
Looks good to me. When you are ready to call for a vote, let us know.

cfep-13.md
```diff
@@ -0,0 +1,70 @@
...
+2. A public, global registry of all outputs from all feedstocks will be kept
+   in a repo on github. Write access to this repo is limited to core and bot 
+   admins as well.
+3. When a feedstock builds new outputs it wants to upload, it will push these outputs
+   to a staging organization on `anaconda.org`. It will also send the names of the
+   outputs, their SHA256 checksums, the feedstock name, and the secret key to a
+   validation web service. This web service then verifies the secret key-feedstock
+   combination and verifies that the outputs are allowed for that feedstock.
+4. If everything is ok, the outputs are copied from the staging organization to the
+   main conda-forge channels.
```
scopatz: We should probably delete bad/old artifacts from the staging channel after some period of time (28 days?).  I don't want the staging channel to be a usuable mirror of conda-forge, or else people could use that and it would be a security hole

isuruf: After copying, we could delete right?

beckermr: Yeah we can do that. But this is on them. They should not use an unsecured channel. We cannot control what they do.

scopatz: I think yeah we could delete after a copy

beckermr: done.

cfep-13.md
scopatz: hmm yeah, I see

cfep-13.md
```diff
@@ -0,0 +1,70 @@
...
+   maintainers of that feedstock and the core team. These keys will be kept in
+   a private repo with only core members having read access. Core members and 
+   encrypted GitHub keys for bot admins will have write access in order to generate 
+   the secret keys and push then to the CI providers.
+2. A public, global registry of all outputs from all feedstocks will be kept
+   in a repo on github. Write access to this repo is limited to core and bot 
+   admins as well.
+3. When a feedstock builds new outputs it wants to upload, it will push these outputs
+   to a staging organization on `anaconda.org`. It will also send the names of the
+   outputs, their SHA256 checksums, the feedstock name, and the secret key to a
```
djsutherland: Are we literally sending the secret key? Better to use public/private keys and sign the message to the web service, right? Just to avoid having the key itself traveling around as much as possible.

beckermr: A lot of services send API keys over https these days (in the authorization bearer headers), so I figured that would be fine here. Are you saying that is not secure enough for this purpose?

scopatz (approved):

beckermr:
@conda-forge/core

This vote will end on 2020-03-11.

mwcraig (approved):

cfep-13.md
djsutherland: I mean, I suppose https is encrypted anyway, and I'm definitely no expert on these things. It just feels more potentially risky in case of accidentally sending requests to the wrong place / etc.

beckermr: Hmmm. At a practical level, either an API key or the private key (to sign messages) needs to be stored on the CI services. For that point of view the risk for spilling it is mostly the same and due to the CI service itself, maintainers making edits etc IIUIC. 

I am not personally familiar with how to store a private key on the CI services or if they are setup to handle such things. Should we put it in an env var?

Unless we have a good answer for how to store a private key, I’m inclined to go the API key route since I don’t think I will mess it up as easily.

pkgw (approved):

cfep-13.md
xhochy: From reading the text, it is not clear to me what the secret key is used for. Is this for signing? Then using async signing should allow us to have a public key for that?

beckermr: The secret key is a standard ApI key. 

Do you know how to put private keys in the CI services?

If we embed an entire file in an env var, will the CO service properly mask it in outputs?

xhochy: You can store private keys in CI services, depends a bit on the service. Travis provides a different public encryption key per repo, for Github Actions you can set it via the UI in Actions Tab (and there is hopefully an API for that), not sure about Azure but I would expect to provide a similar thing. They all will mask the exact value which is in the key in the CI log. The key will also be provided as an env var in the CI job but only when run as part of a commit in `conda-forge`, not in PRs.

beckermr: This entire issue around an API key versus public/private keys is a red herring. 

My main concern is correctly implementing this system both in the code and as a whole with various keys secured correctly. 

I’m assuming I’ll be doing the coding here and I am not confident I can handle the public/private key option correctly. 

A worse outcome is for us to think we have fixed this when there is actually a bug. This is especially true with respect to cryptography. 

If someone else wants to help code this with public/private keys and knows definitely how to secure the private key on our CI services, then I am happy to do that route. 

Otherwise let’s use an API key sent over https since that can be done by me confidently.

beckermr: Setting keys via a UI is a nonstarter. 

It’s not the env var I’m worried about @xhochy. Do you know for sure that an env var with new lines will be masked properly in all of our CI services? Usually people stuff API keys in these, not full files.

xhochy: > It’s not the env var I’m worried about @xhochy. Do you know for sure that an env var with new lines will be masked properly in all of our CI services? Usually people stuff API keys in these, not full files.

With full files, I have no experience, so: no.

beckermr:
@dougalsutherland @xhochy I updated the CFEP to be agnostic between API keys and public/private key pairs. The available people who implement this CFEP can decide when they are working.

xhochy (approved):

ocefpaf (approved):

cfep-13.md
pkgw: If the worry is about multi-line text, it could probably be base64-encoded for env var storage, no?

beckermr: IDK what you are talking about, which is actually the point here. If you have experience doing this correctly and want to proactively help implement this CFEP, then, by all means, lets go for it.

pkgw: Unfortunately I absolutely do not have the spare cycles to help in any concrete way :-( For completeness, [base64](https://en.wikipedia.org/wiki/Base64) encodings can help pack arbitrary binary data into things like environment variables.

beckermr: cool - let's go with the API key approach for now. If you get spare time and have interest, then feel free to pitch in!

jjhelmus (commented):
I've abstaining from this vote due to a possible conflict of interest.  If my vote is needed for this to pass please let me know and I'll re-evaluate.

cfep-13.md
mbargull: `base64` might not be the best choice if the CI would decide to only mask the encoded but not the decoded key :laughing:.
How ever you want to go with it, this would also work:
```sh
# put it in one line:
key_as_one_line="$( < key-file head -n-1 | tail -n+2 | tr -d \\n )"
# reassemble and use:
echo -e "-----BEGIN PRIVATE KEY-----\n${key_as_one_line}\n-----END PRIVATE KEY-----" | \
  openssl dgst -sha256 -sign - some-file > some-file.sha256
```

beckermr:
OK @mbargull I changed the language to be much more generic. LMK!

mbargull:
@beckermr, thanks for the changes, for putting this proposal up, and generally for your work on the infrastructure!

mbargull (approved):

loriab (approved):

msarahan (approved):

CJ-Wright (approved):

ericdill (approved):

minrk (approved):

bgruening (approved):

djsutherland (approved):

mariusvniekerk (approved):

isuruf:
This needs only 12 votes to pass right?

beckermr:
Yup only 12 IIUIC. Since this vote is public, I wanted to give everyone a chance to weigh in either way. However, yes this has passed.

cfep-13.md
```diff
@@ -0,0 +1,64 @@
...
+
+1. A public, global registry of all outputs from all feedstocks will be kept
+   in a repo on github. Write access to this repo is limited to core and bot 
+   admins as well.
+2. When a feedstock builds new outputs it wants to upload, it will push these outputs
+   to a staging organization on `anaconda.org`. These outputs will either be signed 
+   so that they can be validated or the feedstock will directly alert the validation 
+   service in a secure, verifiable manner.
+3. The validation service will validate that any outputs in the staging org are 
+   from the correct feedstock, the actual outputs to be upoloaded, and allowed
```
jakirkham:
```suggestion
   from the correct feedstock, the actual outputs to be uploaded, and allowed 
```

cfep-13.md
```diff
@@ -0,0 +1,64 @@
...
+of say the `python` package. This situation is a security risk that we
+should mitigate.
+
+## Specification
+
+We will secure the uploads of outputs to `anaconda.org` via the following process.
+
+1. A public, global registry of all outputs from all feedstocks will be kept
+   in a repo on github. Write access to this repo is limited to core and bot 
+   admins as well.
```
jakirkham, line 27: Could you please clarify what is being uploaded here? Packages? Something else?

beckermr: this Is an implementation detail and doesn’t belong in the cfep. FWIW I have used json blobs for now.

beckermr: Literally a record of the output names each feedstock generates.

beckermr: Hows that?

jakirkham: Thanks that was what I was looking for 🙂

jakirkham:
It sounds like AppVeyor will need to be dropped as [outlined in this comment]( https://github.com/conda-forge/cfep/pull/26#issuecomment-596840106 ) to enact this proposal. However this presents a challenge with recipes that require MSBuild, which currently only work on AppVeyor (please correct me if I'm missing things). Not sure how we address this issue. Just wanted to raise that concern.

Otherwise agree this would be a nice thing to have.

jakirkham (approved):

mariusvniekerk:
Afaik msbuild things work fine with vs2017

jakirkham:
Ok sounds like we need a separate thing (IDK if it needs to be a CFEP or just a migrator or what) to update to `vs2017` then.

scopatz:
We should leave votes open for the full period, though.

beckermr:
Yes @scopatz this was my plan!

beckermr:
This vote passed!

beckermr:
summary here: https://github.com/conda-forge/conda-forge.github.io/pull/994

PR #32, opened 2020-05-30
cfep-17: Handling pin backports and dependency rebuilds

CJ-Wright:
Thoughts @beckermr?

### DIFF
--- a/cfep-17.md
+++ b/cfep-17.md
@@ -1,0 +1,50 @@
+
+<table>
+<tr><td> Title </td><td> Branches for Globally Pinned Packages </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Christopher J. "CJ" Wright &lt;cjwright4242@gmail.com&gt;</td></tr>
+<tr><td> Created </td><td> May 30, 2020</td></tr>
+<tr><td> Updated </td><td> May 30, 2020</td></tr>
+</table>
+
+## Abstract
+
+The CF global pinnings and the migrations that update the pinnings are important to maintaining consistancy across the CF ecosystem.
+However, the discrepency between the current pin and the version on the master (usually only) branch on the feedstock creates issues.
+
+For instance, if an update needs to be applied to a pinned package (potentially due to a depdendency migration) and the master branch has moved past the pin then a branch needs to be made for the pin.
+This CFEP proposes to automatically create branches for the pinned versions of pinned packages.
+
+This approach would have signifigant benefits:
+1. Make migrations easier for the bot, since the bot can issue PRs into the master and pinned branches
+2. Enable backporting of fixes for pins
+3. Provide the groundwork for backporting of updates to packages
+
+## Specification
+On merge of a PR into the pinned feedstock detect if the new version is beyond its stated compatibility, if so create a branch.
+The bot will only issue migration PRs into master and branch versions which are listed in the pinnings or a migration.
+
+The branches will need a consistent naming, for instance for boost `1.70.x`.
+The `x` acts as a filler for parts of the version number not provided by the `max_pin`.
+
+The bot will open version bump PRs matchign that version spec (for instnace, `1.70.1`) into that branch.
+
+## Other sections
+
+Other relevant sections of the proposal.  Common sections include:
+
+    * Specification -- The technical details of the proposed change.
+    * Motivation -- Why the proposed change is needed.
+    * Rationale -- Why particular decisions were made in the proposal.
+    * Backwards Compatibility -- Will the proposed change break existing
+      packages or workflows.
+    * Alternatives -- Any alternatives considered during the design.
+    * Sample Implementation -- Links to prototype or a sample implementation of
+      the proposed change.
+    * FAQ -- Frequently asked questions (and answers to them).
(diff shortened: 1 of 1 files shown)


cfep-17.md
```diff
@@ -0,0 +1,50 @@
...
+This CFEP proposes to automatically create branches for the pinned versions of pinned packages.
+
+This approach would have signifigant benefits:
+1. Make migrations easier for the bot, since the bot can issue PRs into the master and pinned branches
+2. Enable backporting of fixes for pins
+3. Provide the groundwork for backporting of updates to packages
+
+## Specification
+On merge of a PR into the pinned feedstock detect if the new version is beyond its stated compatibility, if so create a branch.
+The bot will only issue migration PRs into master and branch versions which are listed in the pinnings or a migration.
```
beckermr, line 25: bot behavior doesn't belong here - this cfep is only about automatically making pinning branches

cfep-17.md
```diff
@@ -0,0 +1,50 @@
...
+
+The branches will need a consistent naming, for instance for boost `1.70.x`.
```
beckermr, line 27: i'd suggest something more unique and specified here in the cfep. For example, `cf-pinning-1.70.x` would help users know this is an automated branch.

cfep-17.md
```diff
@@ -0,0 +1,50 @@
...
+The `x` acts as a filler for parts of the version number not provided by the `max_pin`.
+
+The bot will open version bump PRs matchign that version spec (for instnace, `1.70.1`) into that branch.
```
beckermr, line 30: again what the bot does is not relevant here

cfep-17.md
```diff
@@ -0,0 +1,50 @@
...
+The CF global pinnings and the migrations that update the pinnings are important to maintaining consistancy across the CF ecosystem.
+However, the discrepency between the current pin and the version on the master (usually only) branch on the feedstock creates issues.
+
+For instance, if an update needs to be applied to a pinned package (potentially due to a depdendency migration) and the master branch has moved past the pin then a branch needs to be made for the pin.
+This CFEP proposes to automatically create branches for the pinned versions of pinned packages.
+
+This approach would have signifigant benefits:
+1. Make migrations easier for the bot, since the bot can issue PRs into the master and pinned branches
+2. Enable backporting of fixes for pins
+3. Provide the groundwork for backporting of updates to packages
```
beckermr, line 21: We should flesh this out more.

cfep-17.md
isuruf: what about python 3.6 branch?

beckermr: What do you mean? Should we rename it according to this spec? Yes if we want automated updates.

isuruf:
I find it confusing if the bot makes branches on feedstocks I maintain. Here's an alternative proposal,
- A bot makes an issue asking to make a branch
- Maintainer says yes by adding a label
- A github action will create the branch
makes it less confusing

CJ-Wright:
The creation of the branch is an implementation detail, mostly. It is the way by which we can have multiple versions accessible to PR at once. I would say that the maintainer consent is provided by merging the PR that breaks the current pin. Having to respond to an issue seems like a bit of a burden for something most maintainers won't need to worry about.

beckermr:
> The creation of the branch is an implementation detail, mostly.

Right. The issue here is one of consent. 

I am on the fence on this issue and am ok either way.

isuruf:
Whatever you decide here, please don't make it the default for feedstocks that I maintain.

beckermr:
> Whatever you decide here, please don't make it the default for feedstocks that I maintain.

Is opt-out too much to ask, or would you prefer opt-in?

isuruf:
>Is opt-out too much to ask

I'm on 210 teams, so yeah.

CJ-Wright:
Additionally, is the choice a real one? If a maintainer turns down this option, where does that put us, back where we are now? I don't think that is tenable.

beckermr:
> Additionally, is the choice a real one? If a maintainer turns down this option, where does that put us, back where we are now? I don't think that is tenable.

This goes back to the discussions we've been having on keybase. How much power do we exert to  get migrations to go through?

isuruf:
Let's say for example, you create a branch for clang 9, I won't be maintaining it because every time there's a rebuild, I have to build it on the ppc64le server and the aarch64 server. So I won't be merging any PRs unless you commit to doing the work yourself.

beckermr:
Clang 9 is a very hefty example that updating via the bot is honestly not possible. 

A question @isuruf. Do you view people responding to an issue by adding a label as consent for the existence of the branch or as a commitment to maintain that branch too?

I think the difference here matters. We only need the branch to exist so that we can issue prs to it. We do not always need it to be maintained, though we may sometimes.

isuruf:
As a commitment. If there's no branch, people will not try to send patches to these older versions.

cfep-17.md
```diff
@@ -0,0 +1,50 @@
+
+<table>
+<tr><td> Title </td><td> Branches for Globally Pinned Packages </td>
+<tr><td> Status </td><td> Draft </td></tr>
```
beckermr:
```suggestion
<tr><td> Status </td><td> Deferred </td></tr>
```

beckermr:
merging as deferred

PR #33, opened 2020-06-06
[CFEP-18] Uploading binaries from pull requests

chrisburr:
As briefly [discussed on gitter](https://gitter.im/conda-forge/conda-forge.github.io?at=5eda85657f08394226cba840). I'm prepare to implement this if it is accepted.

cc @beckermr  @nehaljwani @stuarteberg


cfep-18.md
```diff
@@ -0,0 +1,60 @@
...
+While it is always possible to build binaries locally it can difficult to set up (e.g. macOS/Windows) or time consuming people only have access to less powerful machines.
+
+## Specification
+
+Upon request, any binaries built in a PR will be uploaded to a file sharing service.
+[Firefox Send](https://github.com/timvisee/ffsend) is proposed in this CFEP however this could change to an alternative service such as [https://transfer.sh/](https://transfer.sh/) or a self hosted equivalent such as [transfer.sh](https://github.com/dutchcoders/transfer.sh) or [0x0](https://github.com/mia-0/0x0).
+
+This change will be implemented by modifying the scripts generated by `conda-smithy` to check for the existence of a file named `CF_UPLOAD_PR_BUILDS` in the recipe directory.
+A similar approach will be used for `staged-recipes`.
+Rerendering will not be required to enable this feature and the `CF_UPLOAD_PR_BUILDS` file will be automatically deleted whenever a feedstock is rerendered.
```
beckermr: Just so I understand, to get the feature in the first place we require at least one rerender. You are saying here that the presence of the file in the feedstock should not affect the CI scripts so that this can be turned on at any time.

chrisburr: Yes

beckermr (changes requested):
Generally, I'd remove most of the implementation details like the name of the file used to signal uploads etc.

chrisburr:
Thanks for the quick review @beckermr. Does this look more appropriate for a CFEP?

cfep-18.md
```diff
@@ -0,0 +1,74 @@
...
+
+Allow package binaries to be uploaded from builds in pull requests.
+
+## Motivation
+
+Sometimes it would be useful to access binaries built in pull requests. Reasons include:
+
+* Testing GUI applications
+* More easily testing the effects on a pull requests on downstream packages. 
+  For example, patches in llvm affecting builds of clang.
```
isuruf, line 22: Shouldn't this be a `test_downstreams`?

isuruf:
Although I appreciate you writing this, I don't see why we need a CFEP for this.

beckermr:
I asked for the CFEP, but if core collectively does not think we need one, then we can implement straight away. The guidance on when we do and do not need a CFEP is a bit thin.

cc @conda-forge/core

isuruf:
We don't want a CFEP for each and every PR to conda-smithy. That's a lot of work to do in part of the PR author and core.

isuruf:
Besides what is proposed doesn't change anything, so I'm fine with no CFEP.. We didn't have a CFEP for most of conda-smithy features implemented lately. CFEP-13 is special because it changed the way we are giving permissions.

beckermr:
I am not proposing CFEPs for every change to smithy OFC. This seemed like a big enough policy shift that a formal way for everyone on core to have their say seemed appropriate. I'm happy to avoid this process too though. Let's close this for now and we can simply make the necessary PRs to smithy.

cfep-18.md
chrisburr: Ideally yes but it's sometimes just impractical with the current infrastructure (like in this case of llvmdev+clangdev or clangdev+root/cling).

PR #37, opened 2020-08-27
Exception section so that we can package things like black

mariusvniekerk:
As a response to https://github.com/conda-forge/staged-recipes/pull/12485#discussion_r478619040 we probably want to adjust this policy so that things we already package adhere to it.

### DIFF
--- a/cfep-05.md
+++ b/cfep-05.md
@@ -92,7 +92,7 @@ NOTE: A rerender needs to happen for these changes to reflect in CI files.
 
 ## Exceptions
 
-Certain packages (for example [black](https://pypi.org/project/black/#history)) follows
+Certain packages (for example [black](https://pypi.org/project/black/#history)) follow
 a release cycle in which they have never had a non-beta/alpha release.  In these cases
 the conda packages for those do *not* need to be published to a prerelease label.
 
--- a/cfep-05.md
+++ b/cfep-05.md
@@ -90,6 +90,15 @@ in `recipe/conda_build_config.yaml` in their respective feedstocks.
 
 NOTE: A rerender needs to happen for these changes to reflect in CI files.
 
+## Exceptions
+
+Certain packages (for example [black](https://pypi.org/project/black/#history)) follows
+a release cycle in which they have never had a non-beta/alpha release.  In these cases
+the conda packages for those do *not* need to be published to a prerelease label.
+
+Once a non-prerelease version of such a package is available and has been merged into the associated feedstock they lose this exception and have to publish prereleases to a prerelease label like other packages.
+
+
 ## Alternatives
 
 * Upload pre-release packages to a different anaconda.org user, perhaps
--- a/cfep-05.md
+++ b/cfep-05.md
@@ -44,7 +44,7 @@ backwards compatibility issues.
 
 ## Specification
 
-Mark the pre-release package on anaconda.org as "dev" or "rc" by adding them to 
+Mark the pre-release package on anaconda.org as "dev" or "rc" by adding them to
 the appropriate label.
 The semantics of these labels should generally follow the
 [guidelines](https://docs.python.org/devguide/devcycle.html#stages) that Python
@@ -96,8 +96,9 @@ Certain packages (for example [black](https://pypi.org/project/black/#history)) 
 a release cycle in which they have never had a non-beta/alpha release.  In these cases
 the conda packages for those do *not* need to be published to a prerelease label.
 
-Once a non-prerelease version of such a package is available and has been merged into the associated feedstock they lose this exception and have to publish prereleases to a prerelease label like other packages.
-
+Once a non-prerelease version of such a package is available and has been merged into
+the associated feedstock they lose this exception and should to publish *future* prereleases
+to a prerelease label like other packages.
 
 ## Alternatives
 
--- a/cfep-05.md
+++ b/cfep-05.md
@@ -97,7 +97,7 @@ a release cycle in which they have never had a non-beta/alpha release.  In these
 the conda packages for those do *not* need to be published to a prerelease label.
 
 Once a non-prerelease version of such a package is available and has been merged into
-the associated feedstock they lose this exception and should to publish *future* prereleases
+the associated feedstock, they lose this exception and should publish *future* prereleases
 to a prerelease label like other packages.
 
 ## Alternatives


cfep-05.md
```diff
@@ -90,6 +90,15 @@ in `recipe/conda_build_config.yaml` in their respective feedstocks.
 
 NOTE: A rerender needs to happen for these changes to reflect in CI files.
 
+## Exceptions
+
+Certain packages (for example [black](https://pypi.org/project/black/#history)) follows
```
beckermr:
```suggestion
Certain packages (for example [black](https://pypi.org/project/black/#history)) follow
```

beckermr (approved):

ocefpaf (approved):

ocefpaf:
We actually already "had that" as a verbal agreement. I never noticed that this was not in the document. Thanks @mariusvniekerk!

kkraus14 (approved):

scopatz (approved):

cfep-05.md
```diff
@@ -90,6 +90,15 @@ in `recipe/conda_build_config.yaml` in their respective feedstocks.
 
 NOTE: A rerender needs to happen for these changes to reflect in CI files.
 
+## Exceptions
+
+Certain packages (for example [black](https://pypi.org/project/black/#history)) follow
+a release cycle in which they have never had a non-beta/alpha release.  In these cases
+the conda packages for those do *not* need to be published to a prerelease label.
+
+Once a non-prerelease version of such a package is available and has been merged into the associated feedstock they lose this exception and have to publish prereleases to a prerelease label like other packages.
```
jakirkham: This could be a bit odd if we are talking about moving prior pre-releases into a special label. Should we clarify that?

mbargull (changes requested):
> they lose this exception and **have to** publish prereleases to a prerelease label

This is stark contrast to the 
> This document outlines how those builds **should** be labeled

in this document.
This should not be merged without addressing possible contradictions.

Also, say you have a software that has
- only `1.*a*` releases in the year 2020,
- `1.*.0` releases in 2021,
- only `2.*a*` releases onward.
  And those `2.*a*` releases are actually usable/decent.

What then? These awkward cases could arise due to social circumstances (confident/in-confident devs or mgmt, loss/gain of maintainers, etc.).
If we add "must"/"have to" wording, we have to be very careful here. This is about **semantics** on naming/categorization that can vary between projects (and even within single projects, given time/changes).

cfep-05.md
mariusvniekerk: we shouldn't move the prior pre-releases.

mariusvniekerk:
> Also, say you have a software that has
> 
> * only `1.*a*` releases in the year 2020,
> * `1.*.0` releases in 2021,
> * only `2.*a*` releases onward.
>   And those `2.*a*` releases are actually usable/decent.

presumably at that point the meaning of a differs.  This is chiefly a concern for calver / zero-ver packages

cfep-05.md
mariusvniekerk: adjusted the wording to refer to future releases

ericdill (approved):

cfep-05.md
jakirkham: Completely agree. Thanks for clarifying that 😄

cfep-05.md
```diff
@@ -90,6 +90,16 @@ in `recipe/conda_build_config.yaml` in their respective feedstocks.
...
 NOTE: A rerender needs to happen for these changes to reflect in CI files.
 
+## Exceptions
+
+Certain packages (for example [black](https://pypi.org/project/black/#history)) follow
+a release cycle in which they have never had a non-beta/alpha release.  In these cases
+the conda packages for those do *not* need to be published to a prerelease label.
+
+Once a non-prerelease version of such a package is available and has been merged into
+the associated feedstock they lose this exception and should to publish *future* prereleases
```
jakirkham: Small grammar nit

```suggestion
the associated feedstock, they lose this exception and should publish *future* prereleases
```

jakirkham (approved):

mariusvniekerk:
@mbargull can you rereview?

scopatz:
Is there going to be a vote called for this?

mbargull (approved):
Sorry, @mariusvniekerk, I didn't notice the edit/request for review before.
With this now consistently saying "should", I'm fine with the amendment.
(I think personally I'd say `## Possible Exceptions` instead of `## Exceptions`, because I didn't ponder about what else might be problematic (and I wouldn't want to indicate the given ones are the only possible exceptions) -- however that would just be me being pedantic+unsure ;D. But the wording is also acceptable as is from my POV.)

mariusvniekerk:
attn: @conda-forge/core

mbargull:
> Is there going to be a vote called for this?

Given that this does not change the mode of suggestion vs prescription, I don't think we need a vote.
(But ppl. should still feel encouraged to review, nontheless!)

scopatz:
I think we should get rid of these "shoulds" as they are causing too much confusion. I read both of these shoulds as a "must", ie with an implicit "if".  "Once [IF] a non-prerelease version of such a package is available ..., [THEN] they lose this exception and should [MUST] publish *future* prereleases"

mariusvniekerk:
i'd view that as out of scope of this particular clarification.

mbargull:
> i'd view that as out of scope of this particular clarification.

I very much agree.

> ... \[MUST\] ...

That change would definitely have to be voted on. Plus, you'd have to do some research and give reasons if such policy would be feasible with the existing diverse software release modes. I'm not saying I am against it, but I'd only support it given enough background information.
If you, however, want to propose another wording that you find less confusion, but which still carries the "should" meaning, we can do that, of course -- I'm definitely supportive on clarifications!

scopatz:
I guess it is fine as is, but I am in favor of not making the problem worse in this PR.  I really do read what is written here as a requirement, so it is disconcerting to me that other there is another, looser interpretation.

ericdill:
On the core dev call today the following was clarified:
* Need to vote to adjust this CFEP as per these changes
* @scopatz will submit a clarifying PR after this one is in.

mariusvniekerk:
@conda-forge/core
This PR falls under the `Overall workflow and packaging policies` policy, please vote and/or comment on this PR.
This PR needs 50% of core to vote yea to pass.
This vote will end on Wednesday September 9, 2020.

scopatz (approved):

ericdill (approved):

pkgw (approved):

CJ-Wright (approved):

mwcraig (approved):

xhochy (approved):

jjhelmus (approved):

marcelotrevisani (approved):

chenghlee (approved):

mariusvniekerk:
Total headcount of core atm is 22 according to https://github.com/conda-forge/conda-forge.github.io/blob/d9a90ce895ab87b9c646b0b02b5322818787a51c/src/core.csv

Thus this vote passes with 12 Yays, and 0 Nays

jakirkham:
Thanks Marius! 😄

PR #42, opened 2021-05-31
Propose to extend clep'05 labels also to Long Term Releases

SrNetoChan:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

### DIFF
--- a/cfep-05.md
+++ b/cfep-05.md
@@ -13,9 +13,10 @@
 
 There are a number of recipe maintainers who have asked for guidance regarding
 how to make pre-release builds available with the CI infrastructure on
-conda-forge. This document outlines how those builds should be labeled on
+conda-forge. On the other hand, some packages have Long Term Releases (LTR) branches
+This document outlines how those builds should be labeled on
 anaconda.org so that the more adventurous users can try out these pre-release
-packages.
+packages and more cautious users can easily use LTR packages.
 
 ## Motivation
 
@@ -44,9 +45,10 @@ backwards compatibility issues.
 
 ## Specification
 
-Mark the pre-release package on anaconda.org as "dev" or "rc" by adding them to
-the appropriate label.
-The semantics of these labels should generally follow the
+Mark the pre-release or long term release package on anaconda.org as "dev" or "rc"
+and "ltr" by adding them to the appropriate label.
+
+The semantics of pre-release labels should generally follow the
 [guidelines](https://docs.python.org/devguide/devcycle.html#stages) that Python
 itself follows.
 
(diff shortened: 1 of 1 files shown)


CJ-Wright:
What does ltr mean in this case? What should users expect when installing from that channel? What kind of support should the maintainers provide? For example are `ltr` labeled packages getting rebuilds of their pinned dependencies?

Edit: respect -> expect typo, sorry

SrNetoChan:
> What does ltr mean in this case? What should users respect when installing from that channel? What kind of support should the maintainers provide? For example are `ltr` labeled packages getting rebuilds of their pinned dependencies?

@CJ-Wright  ltr means long term release. I should have presented my use case: QGIS

In QGIS, we have a normal stable release (3.18 now) and a LTR (3.16 now). They are both stable, but in each 4 stable releases, one becomes an LTR. During an year, the LTR branch only receives fixes and improvements, no new features are allowed.

I maintain a LTR branch on QGIS-feedsock, brasically I just make sure to update the new version and the hash. I also pin a few dependencies so that updates don't risk to break peoples workflows. 

Currently, for the LTR, we teach people to force the QGIS version during the installation. With a label, they could install the LTR without the need to knowing which is the version of the current LTR.

CJ-Wright:
Hmm. I guess my questions are more aimed at what does a long term release mean in the context of conda-forge. This isn't so much an issue with dev/rc labels since they are mostly unsupported, in the sense that if something broke or they weren't rebuilt with the latest global pin it would be fine. My understanding here is that it would be quite different, LTR releases are expected to be supported. Mostly I'm trying to get at if LTR labeled packages have particular expectations attached to them from the conda-forge side, ie rebuilds. If they do, what are they and do we need to make certain they are met?

SrNetoChan:
> Hmm. I guess my questions are more aimed at what does a long term release mean in the context of conda-forge. This isn't so much an issue with dev/rc labels since they are mostly unsupported, in the sense that if something broke or they weren't rebuilt with the latest global pin it would be fine. My understanding here is that it would be quite different, LTR releases are expected to be supported. Mostly I'm trying to get at if LTR labeled packages have particular expectations attached to them from the conda-forge side, ie rebuilds. If they do, what are they and do we need to make certain they are met?

@CJ-Wright, at least for QGIS, the ltr expectations are more or less the same as of the main branch. The reason to maintain an LTR is to have a more stable version available. This is the version the QGIS project suggests for organizations to use in production, that's why I try to keep it available.

Every month there's a new patch release, and I try to get it rebuilt. If something goes wrong, normally I get support from @ocefpaf  or Sam Gillingham. I don't expect any extra involvement from conda-forge like automatic rebuild. I took that task to myself.

I have no idea if there other packages that would benefit or already releasing parallel versions.

isuruf:
Why not have a `qgis-ltr` conda meta package?

SrNetoChan:
@isuruf Sorry, I am not sure how that works.  would that mean a new feedstock? For me, all options are valid. Even keeping the way it is :-) Just want to make it in a way that makes sense for the rest of the conda-forge community.

chrisburr:
The simplest would be to make the lts branch in the feedstock produce multiple outputs, similar to how `matplotlib` is a metapackage around `matplotlib-base`: https://github.com/conda-forge/matplotlib-feedstock/blob/1fb8f9baf8b74041daf9a84a115d00653c9e1cab/recipe/meta.yaml#L75-L87

SrNetoChan:
> The simplest would be to make the lts branch in the feedstock produce multiple outputs, similar to how `matplotlib` is a metapackage around `matplotlib-base`: https://github.com/conda-forge/matplotlib-feedstock/blob/1fb8f9baf8b74041daf9a84a115d00653c9e1cab/recipe/meta.yaml#L75-L87

Seems that this is the preferable approach. I will close this ticket and investigate what you propose. Thank you all for your inputs.

PR #47, opened 2022-07-09
Propose CFEP-21 license packages for static/header libraries

carterbox:
@hmaarrfk, mentioned this in a discussion today, and every so often I wonder about whether licenses for header-only libraries are being satisfied. I wrote something up to see if we can make this official policy.

### DIFF
--- a/cfep-21.md
+++ b/cfep-21.md
@@ -1,0 +1,52 @@
+
+<table>
+<tr><td> Title </td><td> Export license packages for static and header-only libraries </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Daniel Ching &lt;carterbox@users.noreply.github.com&gt;</td></tr>
+<tr><td> Created </td><td> Jul 9, 2022</td></tr>
+<tr><td> Updated </td><td> Jul 9, 2022</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+This CFEP proposes that header-only and static libraries and should be required
+to run_export an empty package (containing their license) called "{{ name
+}}-license" in order to ensure/automate attribution and to satisfy the license
+requirements.
+
+## Specification
+
+Header-only or static library packages should include an output called
+"{{ name }}-license" which is added as a run_export for the library.
+
+## Motivation
+
+Attribution for header-only and static libraries on conda-forge is not well
+enforced because unlike shared libraries which are also required to be
+installed again at runtime (with license included), these libraries need manual
+intervention to ensure that the license of the header-only or static library is
+concatenated to the license metadata of the downstream library which uses it.
+This should obviously be done and is easy to notice for static libraries (whose
+names end in "-static"), but for header-only libraries it is less obvious
+because the name of the library does not obviously mark the library as
+header-only.
+
+Libraries like Eigen and Pybind11 are commonly used, but probably not
+attributed beyond their mention in the host section of the recipe. Mention in
+the requirements section of a recipe probably does not meet the
+licensing/attribution requirements.
+
+## Rationale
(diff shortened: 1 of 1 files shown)


minrk (approved):

beckermr (approved):
Seems good to me. Note that we cannot patch run exports so we won't be able to go back and fix old packages that should have run exports. This is not a big deal but useful to set expectations on how compliant we can be.

hmaarrfk:
>  Seems good to me. Note that we cannot patch run exports so we won't be able to go back and fix old packages that should have run exports. This is not a big deal but useful to set expectations on how compliant we can be.

We can however start a migration. That can likely help this issue.

carterbox:
@conda-forge/core
This vote will end on 2022-08-02 (August 2).

hmaarrfk (approved):

pkgw (approved):

cfep-21.md
```diff
@@ -0,0 +1,60 @@
...
+<tr><td> Author(s) </td><td> Daniel Ching &lt;carterbox@users.noreply.github.com&gt;</td></tr>
+<tr><td> Created </td><td> Jul 9, 2022</td></tr>
+<tr><td> Updated </td><td> Jul 25, 2022</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+This CFEP proposes that header-only and static libraries and should be required
```
bgruening: The second "and" can be removed I think.

carterbox:
```suggestion
This CFEP proposes that header-only and static libraries should be required
```
Thanks!

cfep-21.md
```diff
@@ -0,0 +1,60 @@
...
+<tr><td> Author(s) </td><td> Daniel Ching &lt;carterbox@users.noreply.github.com&gt;</td></tr>
+<tr><td> Created </td><td> Jul 9, 2022</td></tr>
+<tr><td> Updated </td><td> Jul 25, 2022</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+This CFEP proposes that header-only and static libraries should be required
```
isuruf, line 14: Not all licenses require the distribution of the license when the code is distributed in binary form.

carterbox: Thanks for pointing this out. Common licenses like BSD3 and MIT do require distribution of the notice and or license with the binary form though. My thought is that it is easier to require all packages to attribute, than to have reviewers check against a running list of which licenses require attribution.

pkgw: Mmm, I think we should encourage people to be looking at and thinking about the licenses, rather than not. It shouldn't be hard to keep a list of licenses that do/don't have requirements, and if there are gray areas, I think we should be looking into them.

beckermr: I think we state that packages whose licenses require license file distribution must use a run export or other mechanism to do this. 

Others are free to not. There is no reason for us to require more work from folks when the underlying package does not require it.

beckermr (changes requested):
See my comment above.

cfep-21.md
carterbox: >  I think we should encourage people to be looking at and thinking about the licenses

This is a very good point, maintainers should understand the license of their packages, but then why keep a list of licenses that require attribution, that would discourage people from looking at the licenses because they would just look at the list.

> There is no reason for us to require more work from folks when the underlying package does not require it.

As I stated, I think it is more (human) work overall to determine on a case-by-case basis whether a license requires attribution than to just export a license package for every header library. I also don't see any harm in attributing when it is not required.

As a compromise, I propose that header-only libraries without license packages are required to have a comment that the license does not require attribution. Otherwise, it is unclear whether the maintainers have considered this CFEP.

PR #48, opened 2022-07-26
ABI pinning via SONAME in build string

carterbox:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

NOTES:

https://www.debian.org/doc/debian-policy/ch-sharedlibs.html
https://github.com/conda-forge/conda-forge.github.io/issues/157

### DIFF
--- a/cfep-22.md
+++ b/cfep-22.md
@@ -133,7 +133,7 @@ is currently utilized for
 and is common in official linux distributions. It also enables installing
 multiple ABI versions of a library simultaneously (remember to refactor to
 prevent clobbering). The downside of this alternative is that it requires
-multiple recipe outputs which is more difficult to implement).
+multiple recipe outputs which is more difficult to implement.
 
 ```yaml
 {% set name = "libavif" %}
--- a/cfep-22.md
+++ b/cfep-22.md
@@ -127,13 +127,14 @@ upstream package makes any such promises about not breaking the ABI.
 
 Proposed in [this
 issue](https://github.com/conda-forge/conda-forge.github.io/issues/157), this
-alternative renames outputs to `{{ name|lower ~ so_name_major }}`. This method
-is currently utilized for
+alternative produces a new output for the shared library with the name `{{
+name|lower ~ so_name_major }}`. This method is currently utilized for
 [libgfortran](https://github.com/conda-forge/ctng-compilers-feedstock/blob/main/recipe/meta.yaml#L542)
-and is common in official linux distributions. It also enables installing
-multiple ABI versions of a library simultaneously (remember to refactor to
-prevent clobbering). The downside of this alternative is that it requires
-multiple recipe outputs which is more difficult to implement.
+and is common in official linux distributions. It enables installing multiple
+ABI versions of a library simultaneously, but this requires separating the
+headers/docs from the shared libraries to prevent clobbering when multiple ABIs
+are installed in the same environment. This kind of output separation is more
+difficult to implement.
 
 ```yaml
 {% set name = "libavif" %}
@@ -146,7 +147,7 @@ multiple recipe outputs which is more difficult to implement.
 {% set so_name_major = so_version.split('.')[0] %}
 
 package:
-  name: {{ name|lower ~ so_name_major }}
+  name: {{ name|lower }}-splitme
   version: {{ version }}
 
 build:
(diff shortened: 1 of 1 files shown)


cfep-22.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Adding ABI name to package name
+
+This alternatives renames outputs to `{{ name|lower }}{{ so_major_version }}`.
+This is not backward compatible and would require migrating all downstream
+packages to the new output names.
+
+### Prepending ABI to package version
+
+This would mean versioning packages to `{{so_major_version}}.{{verison}}`. This
```
k-dominik: typo: `{{so_major_version}}.{{version}}`

carterbox:
```suggestion
This would mean versioning packages to `{{so_major_version}}.{{version}}`. This
```

hmaarrfk:
I'm not sure where I should report this bug but it seems related https://github.com/conda-forge/hdf5-feedstock/pull/177#issuecomment-1196273311

carterbox:
@hmaarrfk, that lint error has a patch in PR already (but it has not been merged). After the patch the lint will be unaffected by wider use of pinnings with build string matches.

isuruf:
Why not go with the prior art, which is to have a `libgfortran5` output? This is what every other binary distro does.

cfep-22.md
```diff
@@ -0,0 +1,183 @@
...
+always make a new API release for this change, so ABI breaks can be avoided by
+pinning down to the patch version. This is the most conservative approach, but
+is the least flexible. Recipe maintainers can pin to minor API versions if the
+upstream package makes any such promises about not breaking the ABI.
+
+### Adding ABI name to package name
+
+This alternatives renames outputs to `{{ name|lower }}{{ so_major_version }}`.
+This is not backward compatible and would require migrating all downstream
+packages to the new output names.
```
carterbox: > Why not go with the prior art, which is to have a libgfortran5 output? This is what every other binary distro does.

I have addressed this in the Alternatives section. It is also discussed in one of the linked issues: https://github.com/conda-forge/conda-forge.github.io/issues/157

An upside to this approach which I have added is that you can install multiple versions simultaneously.

cfep-22.md
```diff
@@ -0,0 +1,185 @@
...
+always make a new API release for this change, so ABI breaks can be avoided by
+pinning down to the patch version. This is the most conservative approach, but
+is the least flexible. Recipe maintainers can pin to minor API versions if the
+upstream package makes any such promises about not breaking the ABI.
+
+### Adding ABI name to package name
+
+This alternatives renames outputs to `{{ name|lower }}{{ so_major_version }}`.
+This is not backward compatible and would require migrating all downstream
+packages to the new output names. However, it does enable installing multiple
```
isuruf: This is not true. There's a `{{ name|lower }}` output too, so it is backwards compatible. Please see https://github.com/conda-forge/ctng-compilers-feedstock/blob/main/recipe/meta.yaml

carterbox: That's a good point! I'll update this section with a better description and example using your suggestion.

carterbox: I have added a more detailed example of this alternative. I think it is the best alternative. I assume that its major utility is that it allows for multiple ABIs of a library to be installed into the same environment. Judging whether this is an important feature for conda-forge is beyond my depth.

I prefer the build string approach because it is simpler and because I haven't run into a case when I wanted to install multiple versions of the same library. However, this is just my personal experience.

carterbox: Allowing multiple ABI versions of a package simultaneously causes much more work for refactoring a recipe in order to prevent clobbering. I have updated the proposal to better show how splitting a recipe this way would work.

carterbox: Related thought: The purpose of the global conda-forge-pinnings/migrations is to keep all packages in sync with respect to the ABI/API of popular packages. Since publishing ABI-in-the-name packages allows installing multiple ABIs of the same package, does that mean that global conda-forge-pinnings would no longer be needed?

If so, which is easier? Maintaining global pinnings or maintaining individual feedstocks to ensure that ABI-in-the-name packages are implemented (correctly).

isuruf: I'm not suggesting we do multiple ABIs at the same time. We have both `libgfortran` and `libgfortran5` as `run_exports` and if a person really wants multiple ABIs, they can do that, but that's not the default. Because both `libgfortran` and `libgfortran5` are dependencies, you can only have one package.

carterbox:
https://github.com/conda-forge/libidn11-feedstock/blob/main/recipe/meta.yaml

(Another package that uses SONAME in an new output)

hmaarrfk:
I think i would like to see https://github.com/conda-forge/abseil-cpp-feedstock/issues/46 resolved before moving more pinnings to conda build strings. It seems that this would exacerbate the issue

carterbox:
> I think i would like to see [conda-forge/abseil-cpp-feedstock#46](https://github.com/conda-forge/abseil-cpp-feedstock/issues/46) resolved before moving more pinnings to conda build strings. It seems that this would exacerbate the issue

I didn't anticipate that conda was unprepared to do complex matching in the build string. We could end up with some very long build strings:

```
v{{ so_name_major }}so_{{ mpi_prefix }}_{{ cuda_prefix }}_h{{ PKG_HASH }}_{{ build }}
```

which various packages in the same environment could try to match in different ways:

```
foo 1.*     v{{ so_name_major }}so*
foo >=1.1   v{{ so_name_major }}so*{{ mpi_prefix }}*
foo =1.1.*  v{{ so_name_major }}so*{{ cuda_prefix }}*
```
And apparently the conda solver is not equipped to resolve these patterns right now.

h-vetinari:
> And apparently the conda solver is not equipped to resolve these patterns right now.

It will be, as soon https://github.com/conda/conda/pull/11612 is merged (and then percolates into the ecosystem).

carterbox:
Closing in favor of https://github.com/conda-forge/cfep/pull/61

Issue #8, opened 2018-07-17
recipes that are too big to fail


CJ-Wright:
There are core recipes which are "too big to fail". These are recipes, if they are built incorrectly will cause systemic problems across the ecosystem. For example the recent move of some recipes to `noarch` could have caused problems with `conda` itself. 

We need to have tighter control of these recipes.

I think that we need to (at least):
1. Have all (or at least some) of core as maintainers of these packages
1. Require review approval to merge

ericdill:
Can you start putting together a list whose goal is to provide exhaustive coverage of these too big to fail recipes?

CJ-Wright:
Sure, although some help on that front would be good since this can be subjective.

CJ-Wright:
I think we should also put this info in `extra` since this is a property of the feedstock best managed there.

ocefpaf:
I would start with `gdal`, `openvc`, and `qt`.

msarahan:
I disagree with gdal and opencv.  I'm on the fence about qt.

Don't get me wrong, those are important packages.  I don't believe that they are systemic, though.

Python, conda's core deps, the key compression and image libraries (bzip2, zlib, jpeg, libpng, libtiff) are more what I think of for this topic.

ocefpaf:
I used gdal as a guide b/c all of 

> bzip2, zlib, jpeg, libpng, libtiff

are its dependencies. If we can rebuild the stack up to gdal we get roughly 80% of the important packages.

CJ-Wright:
Side note, can we have a core Jinja variable which lists the core developers and then add that to the maintainers for these packages? That way we don't need to explicitly add everyone and as the make-up of core changes these recipes change too. My biggest concern is how conda smithy knows who to add to the feedstock, does it render the recipe?

We'd most likely want to put this variable somewhere central.

ocefpaf:
I like that idea but maybe we may want that to be a subset of the core (maybe volunteers only?).

PR #51, opened 2024-04-12
[CFEP-22] Too Big To Be Compromised (TB2C)

beckermr:
This PR has the "too big to be compromised" proposal from our discussion.

### DIFF
--- a/cfep-22.md
+++ b/cfep-22.md
@@ -1,0 +1,46 @@
+
+<table>
+<tr><td> Title </td><td> TB2C - To Big To Be Compromised </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Matthew R. Becker &lt;becker.mr@gmail.com&gt;</td></tr>
+<tr><td> Created </td><td> April 12, 2024</td></tr>
+<tr><td> Updated </td><td> April 12, 2024</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+There are some feedstocks on `conda-forge` that are "too big to be compromised" (TB2C).  These feedstocks
+form core parts of the ecosystem and a malicious build by a malicious maintainer could have dire consequences
+for our users and the reputation of `conda-forge`.  This CFEP proposes a new policy for these feedstocks to provide
+additional controls on who can build and upload packages from them.
+
+## Specification
+
+The `conda-forge/core` team would maintain a list of TB2C feedstocks in a central, publicly accessible location.
+
+The following rules would be applied to these feedstocks and this list:
+
+- Only members of the `conda-forge/core` team, upstream maintainers of the software in the feedstock, or maintainers
+  approved by the `conda-forge/core` team (either through unanimous agreement of the existing `conda-forge/core` maintainers on
+  the feedstock or with a 50% sensitive vote of the full `conda-forge/core` team) would be allowed access to these feedstocks.
+- Any non-core maintainer can be removed from the feedstock by a 50% sensitive vote of the `conda-forge/core` team.
+- Code reviews are required for all PRs to these feedstocks.
+- No automerge by the autotick bot is allowed for these feedstocks. Automerge labels are allowed.
+- No uploads from outside of the CI (i.e., CFEP-03) are allowed.
+- The full dependency tree of all feedstocks shipping packages the feedstock in question uses in its build is
+  also required to be in the TB2C list.
+- The `conda-forge/core` team can, at its discretion, add or remove feedstocks to/from TB2C list, either through consensus
+  or a 50% public vote if no consensus can be reached.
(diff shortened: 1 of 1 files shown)


jaimergp:
xref https://github.com/conda-forge/cfep/issues/8 for prior discussions

beckermr:
Also see CFEP-08 for a prior proposal. This proposal is largely similar, but specifies voting rules, exempts upstream maintainers to help avoid maintenance issues, allows core maintainers on a feedstock to add folks by unanimous consent, and doesn't specify how we declare the TB2C list except that it is done via consensus and/or a vote.

The goal relative to CFEP-08 is to be less prescriptive in which feedstocks make the list and to spread the maintenance out to trusted community members who may not be on the core team.

cfep-22.md
```diff
@@ -0,0 +1,48 @@
...
+- Any non-core maintainer can be removed from the feedstock by a 50% sensitive vote of the `conda-forge/core` team.
+- Code reviews are required for all PRs to these feedstocks.
+- No automerge by the autotick bot is allowed for these feedstocks. Automerge labels are allowed.
+- No uploads from outside of the CI (i.e., CFEP-03) are allowed.
+- The full dependency tree of all feedstocks shipping packages the feedstock in question uses in its build is
+  also required to be in the TB2C list.
+- The `conda-forge/core` team can, at its discretion, add or remove feedstocks to/from TB2C list, either through consensus
+  or a 50% public vote if no consensus can be reached.
+- Anyone can petition the `conda-forge/core` team to add or remove a feedstock to/from the TB2C list through a 50% public vote.
+- A single vote to add a feedstock to the TB2C list can be used to both add the feedstock and any non-core, non-upstream maintainers
```
jaimergp, line 38: Are we concerned with the delays that this procedure might add to the process? If we need to add a new dependency because it has become part of the protected dependency tree, are we fine with that process taking a couple weeks to be processed?

cfep-22.md
```diff
@@ -0,0 +1,48 @@
...
+The following rules would be applied to these feedstocks and this list:
+
+- Only members of the `conda-forge/core` team, upstream maintainers of the software in the feedstock, or maintainers
+  approved by the `conda-forge/core` team (either through unanimous agreement of the existing `conda-forge/core` maintainers on
+  the feedstock or with a 50% sensitive vote of the full `conda-forge/core` team) would be allowed access to these feedstocks.
+- Any non-core maintainer can be removed from the feedstock by a 50% sensitive vote of the `conda-forge/core` team.
+- Code reviews are required for all PRs to these feedstocks.
+- No automerge by the autotick bot is allowed for these feedstocks. Automerge labels are allowed.
+- No uploads from outside of the CI (i.e., CFEP-03) are allowed.
+- The full dependency tree of all feedstocks shipping packages the feedstock in question uses in its build is
```
jaimergp, line 33: Full dependency tree includes all platforms, I assume.

cfep-22.md
beckermr: I am not concerned about that. We have to make compromises somewhere around security vs keeping the ecosystem open. This seems like one of those places.

cfep-22.md
beckermr: Yes everything.

cfep-22.md
```diff
@@ -0,0 +1,48 @@
...
+The `conda-forge/core` team would maintain a list of TB2C feedstocks in a central, publicly accessible location.
+
+The following rules would be applied to these feedstocks and this list:
+
+- Only members of the `conda-forge/core` team, upstream maintainers of the software in the feedstock, or maintainers
+  approved by the `conda-forge/core` team (either through unanimous agreement of the existing `conda-forge/core` maintainers on
+  the feedstock or with a 50% sensitive vote of the full `conda-forge/core` team) would be allowed access to these feedstocks.
+- Any non-core maintainer can be removed from the feedstock by a 50% sensitive vote of the `conda-forge/core` team.
+- Code reviews are required for all PRs to these feedstocks.
+- No automerge by the autotick bot is allowed for these feedstocks. Automerge labels are allowed.
```
jaimergp: I think this should be formulated in a more generic way. "Automated merges without human intervention are not allowed", in case the autotick bot is phased out or replaced with something else.

beckermr: done!

hmaarrfk:
Do you have a shortlist of packages you are considering? Is "python" (which version) included in these? I would like to run through one to see what dependencies would be affected by one of your "too big to fail" packages.

I feel like many things are "optional" and we have included many "batteries" with our packages which may make this a difficult endeavor.

beckermr:
I do not yet have a list. I want to generate a similar list and see how many maintainers we have who do not satisfy the criterion in the CFEP. We may be in the position of codifying a policy that we de facto already have.

beckermr:
I'm closing this for now since I don't think i'll have time to put energy into it on second thought.

ocefpaf:
> I'm closing this for now since I don't think i'll have time to put energy into it on second thought.

This is an effort that will take some dedication, coordination, and kind helpful comments from all of us. Also, maybe we should write a NF small grant, or some other sort of funding, to help whoever leads this.

Thanks @beckermr for starting it. Let's use this as a base for the next step b/c there are already many good ideas and points here.

PR #52, opened 2024-07-18
CFEP 21 - conda-forge / Bioconda Transfer Policy - DRAFT

mfansler:
I am proposing we formalize transfers between Conda Forge and Bioconda. Conda Forge to Bioconda transfers create problems for other packages on Bioconda that need the newer versions. I am suggesting Conda Forge creates a dedicated label that will be applied to all builds from a transferred feedstock. This would alleviate solving issues by moving the older builds off the `main` label. 

@conda-forge/bioconda-recipes 

## Example Impacted Feedstocks

- https://github.com/conda-forge/r-shazam-feedstock/issues/13
- https://github.com/conda-forge/r-alakazam-feedstock/issues/27
- https://github.com/bioconda/bioconda-recipes/pull/50793

## Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

### DIFF
--- a/cfep-21.md
+++ b/cfep-21.md
@@ -37,7 +37,7 @@ of a new label to be applied to all package builds that transfer CF-to-BC.
 
 ## Motivation
 
-BC package builds use a `strict` channel priority with the channels `conda-forge > bioconda > defaults`. This creates a situation that when a package transfers from CF to BC that the presence of builds with `main` label in `conda-forge` masks the solver from considering builds in lower priority channel. This CFEP proposes to resolve this by moving `conda-forge` builds off the `main` label and onto a dedicated labe. By using a dedicated label for this, we provide end-users with a means of re-enabling search of these older packages by explicitly specifying the label when solving.
+BC package builds use a `strict` channel priority with the channels `conda-forge > bioconda > defaults`. This creates a situation that when a package transfers from CF to BC that the presence of builds with `main` label in `conda-forge` masks the solver from considering builds in lower priority channel. This CFEP proposes to resolve this by moving `conda-forge` builds off the `main` label and onto a dedicated label. By using a dedicated label for this, we provide end-users with a means of re-enabling search of these older packages by explicitly specifying the label when solving.
 
 ## Other sections
 
--- a/cfep-21.md
+++ b/cfep-21.md
@@ -5,8 +5,8 @@
 <tr><td> Author(s) </td><td> Mervin Fansler &lt;mervin.fansler@bric.ku.dk&gt;</td></tr>
 <tr><td> Created </td><td> 10 July 2024</td></tr>
 <tr><td> Updated </td><td> 18 July 2024</td></tr>
-<tr><td> Discussion </td><td> link to the PR where the CFEP is being discussed, NA is circulated initially </td></tr>
-<tr><td> Implementation </td><td> link to the PR for the implementation, NA if not availble </td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda-forge/cfep/pull/52 </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
 </table>
 
 ## Abstract
--- a/cfep-21.md
+++ b/cfep-21.md
@@ -11,7 +11,7 @@
 
 ## Abstract
 
-Bioconda (BC) and Conda Forge (CF) have occasionally transferred package recipes between them. The `strict` channel
+Bioconda (BC) and conda-forge (CF) have occasionally transferred package recipes between them. The `strict` channel
 priority recommended to BC users and used when building packages on the BC channel poses an
 issue when transfers occur, especially in the CF-to-BC direction. This proposal aims to 
 define a policy for standard procedure when such transfers occur. Included in this is the introduction
--- a/cfep-21.md
+++ b/cfep-21.md
@@ -4,7 +4,7 @@
 <tr><td> Status </td><td> Draft </td></tr>
 <tr><td> Author(s) </td><td> Mervin Fansler &lt;mervin.fansler@bric.ku.dk&gt;</td></tr>
 <tr><td> Created </td><td> 10 July 2024</td></tr>
-<tr><td> Updated </td><td> 18 July 2024</td></tr>
+<tr><td> Updated </td><td> 19 September 2024</td></tr>
 <tr><td> Discussion </td><td> https://github.com/conda-forge/cfep/pull/52 </td></tr>
 <tr><td> Implementation </td><td> NA </td></tr>
 </table>
@@ -14,7 +14,7 @@
 Bioconda (BC) and conda-forge (CF) have occasionally transferred package recipes between them. The `strict` channel
 priority recommended to BC users and used when building packages on the BC channel poses an
 issue when transfers occur, especially in the CF-to-BC direction. This proposal aims to 
-define a policy for standard procedure when such transfers occur. Included in this is the introduction
+define a policy for standard procedures when such transfers occur. Included in this is the introduction
 of a new label to be applied to all package builds that transfer CF-to-BC.
 
 ## Specification
@@ -22,22 +22,38 @@ of a new label to be applied to all package builds that transfer CF-to-BC.
 ### Bioconda to conda-forge (BC-to-CF) Transfers
 
 1. Packages on BC considered for transfer should be marked with a ["Move to conda-forge" label](https://github.com/bioconda/bioconda-recipes/labels/Move%20to%20Conda-Forge) in the `bioconda-recipes` repository. This label can be applied either in an Issue or a Pull Request.
-2. A Pull Request should be create on CF's `staged-recipes` repository to add the recipe
+2. A Pull Request should be create on CF's `staged-recipes` repository to add the recipe. A link to the labeled BC Issue/Pull Request should be included in the body of the CF Pull Request.
 3. The linter will flag when adding a recipe already hosted by BC. This will trigger a message to [the **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes).
-4. Upon approval by a **conda-forge/bioconda-recipes** team member, the Pull Request can then be reviewed and merged.
+4. Upon approval by [the **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes), the Pull Request can then be reviewed and merged.
 5. A Pull Request should be created (or edited) on `bioconda-recipes` that removes the recipe. A link to the CF `staged-recipes` Pull Request should be documented in the Pull Request text.
 
 ### conda-forge to Bioconda (CF-to-BC) Transfers
 
-1. An Issue should be created on the CF feedstock with the title "Move to Bioconda". The **conda-forge/bioconda-recipes** team can also be pinged for their input.
-2. A Pull Request should be created on the `conda-forge/admin-requests` feedstock to label all previous builds of this feedstock with a `transferred-to-bioconda` label. It should also initiate a request from the Core team to archive the feedstock. The **conda-forge/bioconda-recipes** should be pinged to review.
+1. An Issue should be created on the CF feedstock with the title "Move to Bioconda". [The **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes) can also be pinged for their input.
+2. A Pull Request should be created on the `conda-forge/admin-requests` feedstock to label all previous builds of this feedstock with a `transferred-to-bioconda` label. It should also initiate a request from the Core team to archive the feedstock. [The **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes) should be pinged to review.
 3. A Pull Request on BC's `bioconda-recipes` should be created to add the recipe there. A link to this Pull Request should be added to the above Issue and Pull Request.
-4. Once a link is provided and a **conda-forge/bioconda-recipes** member has reviewed the Pull Request on `conda-forge/admin-requests`, **conda-forge/core** should proceed to review and merge.
-5. The BC `bioconda-recipes` Pull Request can be merged after **conda-forge/bioconda-recipes** has approved the `conda-forge/admin-requests` Pull Request.
+4. Once a link is provided and a [**conda-forge/bioconda-recipes**](https://github.com/bioconda/bioconda-recipes) member has reviewed the Pull Request on [`conda-forge/admin-requests`](https://github.com/conda-forge/admin-requests), **conda-forge/core** should proceed to review and merge.
+5. The BC `bioconda-recipes` Pull Request can be merged after [the **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes) has approved the `conda-forge/admin-requests` Pull Request.
+
+### Expected Implementation
+
+1. An new option `cfep21_label` will be added to [`conda-forge/admin-requests`](https://github.com/conda-forge/admin-requests).
+2. Once implemented, the following information will be added to the conda-forge Documentation:
+   - A section on BC-to-CF transfer will be added under "Maintainer Documentation > Contributing packages".
+   - A section on CF-to-BC transfer will be added under "Maintainer Documentation > Maintaining packages".
+   - A section on using the `label/transferred-to-bioconda` label to solve environments requiring the labelled artifacts will be added under "User Documentation > FAQ".
 
 ## Motivation
 
-BC package builds use a `strict` channel priority with the channels `conda-forge > bioconda > defaults`. This creates a situation that when a package transfers from CF to BC that the presence of builds with `main` label in `conda-forge` masks the solver from considering builds in lower priority channel. This CFEP proposes to resolve this by moving `conda-forge` builds off the `main` label and onto a dedicated label. By using a dedicated label for this, we provide end-users with a means of re-enabling search of these older packages by explicitly specifying the label when solving.
+BC package builds use a `strict` channel priority with the channels `conda-forge > bioconda > defaults`. This creates a situation that when a package transfers from CF to BC the presence of builds with `main` label in `conda-forge` masks the solver from considering builds in a lower priority channel.
+
+## Rationale
+
+This CFEP proposes to resolve CF-to-BC transfer issues by moving `conda-forge` builds off the `main` label and onto a dedicated label. This approach relies on a metadata update to hide the artifacts in `conda-forge` channel so that artifacts in `bioconda` channel become visible in `strict` channel priority. By using a dedicated label, end-users will also have a means of re-enabling search of these older packages by specifying the label when solving.
+
+## Alternatives
+
+Transferring artifacts (e.g., as in [CFEP-3](https://github.com/conda-forge/cfep/blob/main/cfep-03.md), followed by removal) would also solve the end-user issues. However, this would can be regarded as a more drastic change compared to a labeling approach.
 
 ## Other sections
 
--- a/cfep-21.md
+++ b/cfep-21.md
@@ -2,7 +2,7 @@
 <table>
 <tr><td> Title </td><td> Conda Forge-Bioconda Transfer Policy </td>
 <tr><td> Status </td><td> Draft </td></tr>
-<tr><td> Author(s) </td><td> Mervin Fansler &lt;fanslerm@mskcc.org&gt;</td></tr>
+<tr><td> Author(s) </td><td> Mervin Fansler &lt;mervin.fansler@bric.ku.dk&gt;</td></tr>
 <tr><td> Created </td><td> 10 July 2024</td></tr>
 <tr><td> Updated </td><td> 10 July 2024</td></tr>
 <tr><td> Discussion </td><td> link to the PR where the CFEP is being discussed, NA is circulated initially </td></tr>
--- a/cfep-21.md
+++ b/cfep-21.md
@@ -1,6 +1,6 @@
 
 <table>
-<tr><td> Title </td><td> Conda Forge-Bioconda Transfer Policy </td>
+<tr><td> Title </td><td> conda-forge / Bioconda Transfer Policy </td>
 <tr><td> Status </td><td> Draft </td></tr>
 <tr><td> Author(s) </td><td> Mervin Fansler &lt;mervin.fansler@bric.ku.dk&gt;</td></tr>
 <tr><td> Created </td><td> 10 July 2024</td></tr>
--- a/cfep-21.md
+++ b/cfep-21.md
@@ -21,7 +21,7 @@ of a new label to be applied to all package builds that transfer CF-to-BC.
 
 ### Bioconda to conda-forge (BC-to-CF) Transfers
 
-1. Packages on BC considered for transfer should be marked with a ["Move to Conda Forge" label](https://github.com/bioconda/bioconda-recipes/labels/Move%20to%20Conda-Forge) in the `bioconda-recipes` repository. This label can be applied either in an Issue or a Pull Request.
+1. Packages on BC considered for transfer should be marked with a ["Move to conda-forge" label](https://github.com/bioconda/bioconda-recipes/labels/Move%20to%20Conda-Forge) in the `bioconda-recipes` repository. This label can be applied either in an Issue or a Pull Request.
 2. A Pull Request should be create on CF's `staged-recipes` repository to add the
 3. The linter will flag when adding a recipe already hosted by BC. This will trigger a message to [the **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes).
 4. Upon approval by a **conda-forge/bioconda-recipes** team member, the Pull Request can then be reviewed and merged.
--- a/cfep-21.md
+++ b/cfep-21.md
@@ -1,0 +1,46 @@
+
+<table>
+<tr><td> Title </td><td> Conda Forge-Bioconda Transfer Policy </td>
+<tr><td> Status </td><td> Draft | Proposed | Accepted | Rejected | Deferred | Implemented </td></tr>
+<tr><td> Author(s) </td><td> Mervin Fansler &lt;fanslerm@mskcc.org&gt;</td></tr>
+<tr><td> Created </td><td> Jan 03, 2024</td></tr>
+<tr><td> Updated </td><td> Jan 03, 2024</td></tr>
+<tr><td> Discussion </td><td> link to the PR where the CFEP is being discussed, NA is circulated initially </td></tr>
+<tr><td> Implementation </td><td> link to the PR for the implementation, NA if not availble </td></tr>
+</table>
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


cfep-21.md
```diff
@@ -0,0 +1,60 @@
+
+<table>
+<tr><td> Title </td><td> Conda Forge-Bioconda Transfer Policy </td>
```
jaimergp:
```suggestion
<tr><td> Title </td><td> conda-forge / Bioconda Transfer Policy </td>
```

We style conda-forge as such, lowercase, with dash. Same elsewhere. I added the slash to avoid the two dashes.

cfep-21.md
```diff
@@ -0,0 +1,60 @@
...
+<tr><td> Author(s) </td><td> Mervin Fansler &lt;mervin.fansler@bric.ku.dk&gt;</td></tr>
+<tr><td> Created </td><td> 10 July 2024</td></tr>
+<tr><td> Updated </td><td> 18 July 2024</td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda-forge/cfep/pull/52 </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+Bioconda (BC) and Conda Forge (CF) have occasionally transferred package recipes between them. The `strict` channel
```
jaimergp:
```suggestion
Bioconda (BC) and conda-forge (CF) have occasionally transferred package recipes between them. The `strict` channel
```

cfep-21.md
```diff
@@ -0,0 +1,60 @@
...
+priority recommended to BC users and used when building packages on the BC channel poses an
+issue when transfers occur, especially in the CF-to-BC direction. This proposal aims to 
+define a policy for standard procedure when such transfers occur. Included in this is the introduction
+of a new label to be applied to all package builds that transfer CF-to-BC.
+
+## Specification
+
+### Bioconda to Conda Forge (BC-to-CF) Transfers
```
jaimergp:
```suggestion
### Bioconda to conda-forge (BC-to-CF) Transfers
```

cfep-21.md
```diff
@@ -0,0 +1,60 @@
...
+
+1. Packages on BC considered for transfer should be marked with a ["Move to Conda Forge" label](https://github.com/bioconda/bioconda-recipes/labels/Move%20to%20Conda-Forge) in the `bioconda-recipes` repository. This label can be applied either in an Issue or a Pull Request.
```
jaimergp:
```suggestion
1. Packages on BC considered for transfer should be marked with a ["Move to conda-forge" label](https://github.com/bioconda/bioconda-recipes/labels/Move%20to%20Conda-Forge) in the `bioconda-recipes` repository. This label can be applied either in an Issue or a Pull Request.
```

cfep-21.md
```diff
@@ -0,0 +1,60 @@
...
+2. A Pull Request should be create on CF's `staged-recipes` repository to add the
```
jaimergp: This seems incomplete?

cfep-21.md
```diff
@@ -0,0 +1,60 @@
...
+3. The linter will flag when adding a recipe already hosted by BC. This will trigger a message to [the **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes).
+4. Upon approval by a **conda-forge/bioconda-recipes** team member, the Pull Request can then be reviewed and merged.
```
jaimergp: I guess we should get sign off from bioconda members in this CFEP too. cc @conda-forge/bioconda-recipes

cfep-21.md
```diff
@@ -0,0 +1,60 @@
...
+5. A Pull Request should be created (or edited) on `bioconda-recipes` that removes the recipe. A link to the CF `staged-recipes` Pull Request should be documented in the Pull Request text.
+
+### Conda Forge to Bioconda (CF-to-BC) Transfers
```
jaimergp:
```suggestion
### conda-forge to Bioconda (CF-to-BC) Transfers
```

cfep-21.md
```diff
@@ -0,0 +1,60 @@
...
+
+1. An Issue should be created on the CF feedstock with the title "Move to Bioconda". The **conda-forge/bioconda-recipes** team can also be pinged for their input.
+2. A Pull Request should be created on the `conda-forge/admin-requests` feedstock to label all previous builds of this feedstock with a `transferred-to-bioconda` label. It should also initiate a request from the Core team to archive the feedstock. The **conda-forge/bioconda-recipes** should be pinged to review.
```
jaimergp: We might need to add code for this.

But I do see one issue: the package will disappear from main for those that do not use bioconda. What's the recommended practice here? Them adding the label after experience breakage? Is there a way to avoid the breakage from happening?

cfep-21.md
```diff
@@ -0,0 +1,60 @@
...
+3. A Pull Request on BC's `bioconda-recipes` should be created to add the recipe there. A link to this Pull Request should be added to the above Issue and Pull Request.
+4. Once a link is provided and a **conda-forge/bioconda-recipes** member has reviewed the Pull Request on `conda-forge/admin-requests`, CF Core should proceed to review and merge.
```
jaimergp:
```suggestion
4. Once a link is provided and a **conda-forge/bioconda-recipes** member has reviewed the Pull Request on `conda-forge/admin-requests`, **conda-forge/core** should proceed to review and merge.
```

cfep-21.md
```diff
@@ -0,0 +1,60 @@
...
+5. The BC `bioconda-recipes` Pull Request can be merged after **conda-forge/bioconda-recipes** has approved the `conda-forge/admin-requests` Pull Request.
+
+## Motivation
+
+BC package builds use a `strict` channel priority with the channels `conda-forge > bioconda > defaults`. This creates a situation that when a package transfers from CF to BC that the presence of builds with `main` label in `conda-forge` masks the solver from considering builds in lower priority channel. This CFEP proposes to resolve this by moving `conda-forge` builds off the `main` label and onto a dedicated labe. By using a dedicated label for this, we provide end-users with a means of re-enabling search of these older packages by explicitly specifying the label when solving.
```
jaimergp:
```suggestion
BC package builds use a `strict` channel priority with the channels `conda-forge > bioconda > defaults`. This creates a situation that when a package transfers from CF to BC that the presence of builds with `main` label in `conda-forge` masks the solver from considering builds in lower priority channel. This CFEP proposes to resolve this by moving `conda-forge` builds off the `main` label and onto a dedicated label. By using a dedicated label for this, we provide end-users with a means of re-enabling search of these older packages by explicitly specifying the label when solving.
```

jaimergp:
I wasn't aware that conda-forge recipes were being migrated to bioconda. I would have expected that a better solution was to onboard maintainers to the (I presume) outdated feedstocks, and this way we don't have to deal with these priority issues.

Is this extended practice or just a new case found in this `r-alakazam` package?

cfep-21.md
mfansler: Yes, users would experience an inability to find the package and would have to add the label to use it again. I don't think there is a way around this. It would be discoverable on Anaconda Cloud web searching, but perhaps something should be added to the archived feedstock itself. Can a message be added to the README? or create a pinned Issue with a link to documentation?

Maybe I can put some numbers together on what packages actually have moved before, but in my experience they are already bioinformatics oriented, so my expectation is that most users of such packages will already have Bioconda configured.

mfansler:
> I wasn't aware that conda-forge recipes were being migrated to bioconda. I would have expected that a better solution was to onboard maintainers to the (I presume) outdated feedstocks, and this way we don't have to deal with these priority issues.
> 
> Is this extended practice or just a new case found in this `r-alakazam` package?

I can try to add more examples at some point. The ones I see are driven by R packages adding a Bioconductor dependency, none of which we host. It is acutely problematic for `r-alakazam` because another Bioconda package has a lower bounds dependency above any of the CF builds.

An alternative solution could be that Bioconda creates an ability for recipes to optionally build with `flexible` channel priority. However, I still don't think it would hurt to formalize a transfer policy.

jaimergp:
Ah, I see, not a matter of lack of maintenance, but new requirements not found in conda-forge and available in bioconda. I see.

Even if you manage to build with flexible, downstream users would have to also realize that, and they might be using strict, so yes, I don't think there's a way around 😬 

If only conda could tell you why your package disappeared... I envision a `removed` section in repodata.json where not just the package is listed (like it is now), but there's a reason attached to it too, and we can display it to the user.

beckermr (changes requested):
Has this been discussed with the bioconda dev folks?

mfansler:
> Has this been discussed with the bioconda dev folks?

No. Hopefully they can weigh in here.

CC: @conda-forge/bioconda-recipes

cfep-21.md
bgruening:
```suggestion
2. A Pull Request should be create on CF's `staged-recipes` repository to add the recipe
```

bgruening (commented):
Looks good to me. Thanks @mfansler for writing this up.

I have one question for the strict-channel setting. If I have package A-0.1 in CF and A-0.5 in BC. Is a pin `- A >=0.5` not working? 

The other theoretical option that we have is to transfer the entire repo from CF from BC in anaconda.org. Ideally we could copy the repo (change the labels on the CF side), but I think this is not possible.

mfansler:
@bgruening sorry for the delay in this. I'll try to touch it up to move it from Draft to Proposed.

> I have one question for the strict-channel setting. If I have package A-0.1 in CF and A-0.5 in BC. Is a pin `- A >=0.5` not working?

No, it does not work. In `strict` mode, having any version of package A in CF has the solver ignore all repodata in BC about A.

mfansler:
@bgruening @conda-forge/bioconda-recipes I've updated this. Please review.

Note that the `bioconda-recipes` linter seems to strictly reject CF-to-BC transfers currently, such that these transfers cannot proceed further in PR CI without some solution along the proposed lines. (e.g., https://github.com/bioconda/bioconda-recipes/pull/50793)

@beckermr is it necessary to prototype an implementation (e.g., a PR to `admin-requests`) to move this from "Draft" to "Proposal"?

cfep-21.md
mfansler: Are you thinking a specific number? @bgruening and I are both members.

cfep-21.md
mfansler: I mention in the Implementation section that we would add a section under "User Documentation > FAQ" to cover how to use the label.

jakirkham:
cc @mbargull

ggabernet:
Hi @mfansler, thanks for looking into that. Is there any progress here, as the issue with our two recipes still persists.

ggabernet:
Hi @mfansler @beckermr @conda-forge/core is there a way to move this forward? It would be great to be able to solve this with our packages :)

mfansler:
Last I checked in on this (January), there was openness in Bioconda to enable recipes to override strict channel priority to use flexible solve instead. In the end, that would likely be a more simple implementation, but I haven't yet looked into what work would need to be done to make this possible.

beckermr:
We need to double check support for "flexible" solver priority across mamba and rattler. We may not have it in all cases.

ggabernet:
@mfansler @beckermr I see that it would be advantageous to rather archive the old conda-forge recipes than allow flexible solve within bioconda, as also @jaimergp pointed out "even if you manage to build with flexible, downstream users would have to also realize that, and they might be using strict". They would then be using a very outdated version.

ggabernet:
@beckermr @mfansler just bumping this as we'd still very much like to get a conda-installable up to date version of our packages

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+    * Motivation -- Why the proposed change is needed.
+    * Rationale -- Why particular decisions were made in the proposal.
+    * Backwards Compatibility -- Will the proposed change break existing
+      packages or workflows.
+    * Alternatives -- Any alternatives considered during the design.
+    * Sample Implementation -- Links to prototype or a sample implementation of
+      the proposed change.
+    * FAQ -- Frequently asked questions (and answers to them).
+    * Resolution -- A short summary of the decision made by the community.
+    * Reference -- Any references used in the design of the CFEP.
```
jaimergp, line 72:
```suggestion
```

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+priority recommended to BC users and used when building packages on the BC channel poses an
+issue when transfers occur, especially in the CF-to-BC direction. This proposal aims to 
+define a policy for standard procedures when such transfers occur. Included in this is the introduction
+of a new label to be applied to all package builds that transfer CF-to-BC.
+
+## Specification
+
+### Bioconda to conda-forge (BC-to-CF) Transfers
+
+1. Packages on BC considered for transfer should be marked with a ["Move to conda-forge" label](https://github.com/bioconda/bioconda-recipes/labels/Move%20to%20Conda-Forge) in the `bioconda-recipes` repository. This label can be applied either in an Issue or a Pull Request.
```
jaimergp, line 24: I'd suggest not using too specific terms or guidelines in the Specification, in case those specifics change over time. For example, I'd remove the label details here, and then add an "Example workflow" section where those details are mentioned. The Specification should only say "Create an issue in `bioconda/bioconda-recipes` to notify the intent of transferring a package to conda-forge".

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+2. A Pull Request should be create on CF's `staged-recipes` repository to add the recipe. A link to the labeled BC Issue/Pull Request should be included in the body of the CF Pull Request.
```
jaimergp, line 25: This should only happen once the bioconda team has agreed to allow the transfer to minimize noise. The reference to the comment in the bioconda org allowing the transfer should be mandatory.

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+3. The linter will flag when adding a recipe already hosted by BC. This will trigger a message to [the **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes).
```
jaimergp, line 26: This is an implementation detail we don't need in the specification. Perhaps in the Example workflow section.

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+4. Upon approval by [the **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes), the Pull Request can then be reviewed and merged.
```
jaimergp, line 27: This is implicit if the suggestions to point 2 are applied.

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+5. A Pull Request should be created (or edited) on `bioconda-recipes` that removes the recipe. A link to the CF `staged-recipes` Pull Request should be documented in the Pull Request text.
+
+### conda-forge to Bioconda (CF-to-BC) Transfers
+
+1. An Issue should be created on the CF feedstock with the title "Move to Bioconda". [The **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes) can also be pinged for their input.
```
jaimergp, line 32: Same comment here about specifics, e.g. the title.

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+2. A Pull Request should be created on the `conda-forge/admin-requests` feedstock to label all previous builds of this feedstock with a `transferred-to-bioconda` label. It should also initiate a request from the Core team to archive the feedstock. [The **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes) should be pinged to review.
```
jaimergp, line 33: This should only happen _after_ explicit approval by the feedstock maintainers and the bioconda team. I think I'm inclined to mark them as broken instead of relabeling. Not sure what happens with the original URL once they are relabeled, but we want to avoid 404s in lockfiles.

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+3. A Pull Request on BC's `bioconda-recipes` should be created to add the recipe there. A link to this Pull Request should be added to the above Issue and Pull Request.
```
jaimergp, line 34: This should happen before the admin-requests PR. Once approved (but before the merge), then the admin-requests PR should be submitted. Both PRs should be merged as simultaneously as possible.

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+4. Once a link is provided and a [**conda-forge/bioconda-recipes**](https://github.com/bioconda/bioconda-recipes) member has reviewed the Pull Request on [`conda-forge/admin-requests`](https://github.com/conda-forge/admin-requests), **conda-forge/core** should proceed to review and merge.
```
jaimergp, line 35: We need an extra step here where all the conda-forge packages for that named are copied to bioconda. I think this can be done programmatically with little effort.

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+5. The BC `bioconda-recipes` Pull Request can be merged after [the **conda-forge/bioconda-recipes** team](https://github.com/orgs/conda-forge/teams/bioconda-recipes) has approved the `conda-forge/admin-requests` Pull Request.
+
+### Expected Implementation
```
jaimergp, line 38:
```suggestion
### Recommended Implementation
```

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. An new option `cfep21_label` will be added to [`conda-forge/admin-requests`](https://github.com/conda-forge/admin-requests).
+2. Once implemented, the following information will be added to the conda-forge Documentation:
+   - A section on BC-to-CF transfer will be added under "Maintainer Documentation > Contributing packages".
+   - A section on CF-to-BC transfer will be added under "Maintainer Documentation > Maintaining packages".
+   - A section on using the `label/transferred-to-bioconda` label to solve environments requiring the labelled artifacts will be added under "User Documentation > FAQ".
```
jaimergp, line 44: I don't think we need to discuss the documentation sections that explicitly.

jaimergp (changes requested):
Sorry for dropping this, adding this to my list!

I think we need a clearer discussion on how this will impact end-users that only want to install packages and don't care where they come from. They should not need to change the configured channel priority and, at most, add a new channel to the list in a way that doesn't impact the environment resolution. This should be added to Rationale. Example cases I'd like to see discussed:

- How a conda-forge user may be able to install the package that moved to conda-forge. e.g. add `bioconda` to the channel list in which position and with which priority.
- How a bioconda user may use the package that moved to conda-forge. Probably no changes, because it's already part of their config.

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+
+BC package builds use a `strict` channel priority with the channels `conda-forge > bioconda > defaults`. This creates a situation that when a package transfers from CF to BC the presence of builds with `main` label in `conda-forge` masks the solver from considering builds in a lower priority channel.
+
+## Rationale
+
+This CFEP proposes to resolve CF-to-BC transfer issues by moving `conda-forge` builds off the `main` label and onto a dedicated label. This approach relies on a metadata update to hide the artifacts in `conda-forge` channel so that artifacts in `bioconda` channel become visible in `strict` channel priority. By using a dedicated label, end-users will also have a means of re-enabling search of these older packages by specifying the label when solving.
+
+## Alternatives
+
+Transferring artifacts (e.g., as in [CFEP-3](https://github.com/conda-forge/cfep/blob/main/cfep-03.md), followed by removal) would also solve the end-user issues. However, this would can be regarded as a more drastic change compared to a labeling approach.
```
jaimergp, line 56: I like this, tbh.

cfep-21.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. An new option `cfep21_label` will be added to [`conda-forge/admin-requests`](https://github.com/conda-forge/admin-requests).
+2. Once implemented, the following information will be added to the conda-forge Documentation:
+   - A section on BC-to-CF transfer will be added under "Maintainer Documentation > Contributing packages".
+   - A section on CF-to-BC transfer will be added under "Maintainer Documentation > Maintaining packages".
+   - A section on using the `label/transferred-to-bioconda` label to solve environments requiring the labelled artifacts will be added under "User Documentation > FAQ".
+
+## Motivation
+
+BC package builds use a `strict` channel priority with the channels `conda-forge > bioconda > defaults`. This creates a situation that when a package transfers from CF to BC the presence of builds with `main` label in `conda-forge` masks the solver from considering builds in a lower priority channel.
```
jaimergp, line 48: Why is conda-forge placed before `bioconda`?

cfep-21.md
dlaehnemann: Yes, this sounds like a viable alternative.

cfep-21.md
dlaehnemann: Because previously, packages have only migrated from bioconda to conda-forge, especially CRAN packages that don't have any bioconda dependencies. To properly resolve them, the channel order is recommended in that way. At least I think that was the motivation, but I gladly stand corrected by anybody with more in-depth knowledge.

dlaehnemann:
I ran into the same issue, a CRAN package adding a `bioconductor` dependency for a `bioconductor` package that lives on `bioconda`. By the established rules, this suggests to migrate this package over to `bioconda`, but this is a headache with the current setup.

Just to cross-reference in my attempts, here is [the attempt to create an `r-nmf` package on `bioconda`](https://github.com/bioconda/bioconda-recipes/pull/64030#issuecomment-4165056339), which fails because of the `channel_priority: strict`, with the existing old package versions on `conda-forge`. And here is the [first failing version of `r-nmf` on `conda-forge`, version `0.22.0`](https://github.com/conda-forge/r-nmf-feedstock/pull/6#issuecomment-585206923).

Looking into this repeatedly, I think the two viable options I see would be the following. I would prefer the first one over the second, but wanted to put this up for reviving the discussion here:

1. Simply leave the package on `conda-forge`, [adding the `bioconda` channel to the `recipe/conda_build_config.yaml`](https://github.com/conda-forge/gtdb_validation_tk-feedstock/blob/a51adf979bdb5bd900d1ba564505cd8b9e4fac6f/recipe/conda_build_config.yaml#L2) and dropping windows builds with [`skip: true # [win]`](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#skipping-builds), as bioconda doesn't build for windows. This at least builds `r-nmf` again for linux and mac, for all of the numerous packages that depend on it, especially the `bioconductor` ones. And it doesn't cause any confusion with the discoverability and priority across channels. And moving the package to `bioconda` would also drop windows builds, anyways...
2. Go ahead with this migration proposal here, and [transfer the old `conda-forge` build artefacts over to `bioconda` manually, as suggested as an alternative here](https://github.com/conda-forge/cfep/pull/52#discussion_r2464053486). I think this would be the cleanest solution from the user perspective. The only thing that it breaks. is when users have environment definitions that only include `conda-forge` and not `bioconda` as a channel, but still want to install `r-nmf`.

beckermr:
> I ran into the same issue, a CRAN package adding a `bioconductor` dependency for a `bioconductor` package that lives on `bioconda`. By the established rules, this suggests to migrate this package over to `bioconda`, but this is a headache with the current setup.
> 
> Just to cross-reference in my attempts, here is [the attempt to create an `r-nmf` package on `bioconda`](https://github.com/bioconda/bioconda-recipes/pull/64030#issuecomment-4165056339), which fails because of the `channel_priority: strict`, with the existing old package versions on `conda-forge`. And here is the [first failing version of `r-nmf` on `conda-forge`, version `0.22.0`](https://github.com/conda-forge/r-nmf-feedstock/pull/6#issuecomment-585206923).
> 
> Looking into this repeatedly, I think the two viable options I see would be the following. I would prefer the first one over the second, but wanted to put this up for reviving the discussion here:
> 
> 1. Simply leave the package on `conda-forge`, [adding the `bioconda` channel to the `recipe/conda_build_config.yaml`](https://github.com/conda-forge/gtdb_validation_tk-feedstock/blob/a51adf979bdb5bd900d1ba564505cd8b9e4fac6f/recipe/conda_build_config.yaml#L2) and dropping windows builds with [`skip: true # [win]`](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#skipping-builds), as bioconda doesn't build for windows. This at least builds `r-nmf` again for linux and mac, for all of the numerous packages that depend on it, especially the `bioconductor` ones. And it doesn't cause any confusion with the discoverability and priority across channels. And moving the package to `bioconda` would also drop windows builds, anyways...
> 2. Go ahead with this migration proposal here, and [transfer the old `conda-forge` build artefacts over to `bioconda` manually, as suggested as an alternative here](https://github.com/conda-forge/cfep/pull/52#discussion_r2464053486). I think this would be the cleanest solution from the user perspective. The only thing that it breaks. is when users have environment definitions that only include `conda-forge` and not `bioconda` as a channel, but still want to install `r-nmf`.

So a third, somewhat simpler proposal that would not require a CFEP would be to:

- copy the artifacts from conda-forge to bioconda
- mark them broken on conda-forge (a special procedure for conda-forge that removes them from the main label repodata, but keeps them available for lockfiles, explicit env exports, etc.)
- add the package to bioconda
- change the conda-forge recipe by having it be a simple metapackage that installs from bioconda via `bioconda::<package>`
- possibly backfill old package versions with metapackages as well

I think this would: 

- fix the solver priority issue above (solver priority is enforced for the items found in the repodata at run time IIUIC)
- ensure that folks using only conda-forge can get at least get some sort of solver problem or issue or even an install of the package for updated versions
- ensure that folks using both bioconda and conda-forge can get updated versions
- avoid a circular dependence in the conda-forge feedstock on using bioconda to build a package (instead it is only used to point to bioconda via the metapackage, which is less fragile)

Thoughts?

dlaehnemann:
Yes, this actually does sound like it seamlessly solves the all issues for users. Thanks for thinking this through with the extra insights you have into conda-forge workings and capabilities. I wouldn't have known about the broken label, or the ability to backfill old package versions.

Only the migration process is still a bit lengthy, with multiple steps. So maybe it does make sense to update this CFEP with all of these steps, fleshing out how exactly they are done and who will do what? Is there anything I can do to help with this?

beckermr:
Let's try it once and see how it goes. If it actually works, then I think we add the steps to the docs and call it good. CFEPs are about policy, and this stuff is all about details.

dlaehnemann:
Ah, OK. The docs probably are more discoverable, as well. Should we try it out with `r-nmf`? I would be up for getting started on this.

The only caveat is, that I will be starting parental leave any day now, and would then be away for 6 weeks. But until then I would make this a priority. And the [`bioconda` pull request for `r-nmf` already exists](https://github.com/bioconda/bioconda-recipes/pull/64030), so as I don't have any permissions for `conda-forge`, most stuff would have to be handled by others, anyway.

Also, to give a little context: The package version I propose for `bioconda` is an old one, but the first one that didn't get built on `conda-forge`. So the idea is to add this and then go through all the incremental updates version by version, so we have all the newer versions up on `bioconda` through the regular process.

beckermr:
Yes we can start with this package for sure. Let's make a tracking issue and try to get as far as we can when folks have time. You should definitely not worry about this though. Lots of changes coming your way!

beckermr:
See https://github.com/conda-forge/conda-forge.github.io/issues/2809. As we finish steps, we can mark them as done there.

PR #54, opened 2024-09-20
CFEP 23: Zulip migration

jaimergp:
As discussed internally, let's make a decision about this. I'll open an initial period of discussion of two weeks, but if there's good consensus before then, I won't wait before calling the vote.

If you are a @conda-forge/core member and you are reading this for the first time, don't hesitate to ask for more information. Invite links are available in the Element core room upon request.

### DIFF
--- a/cfep-23.md
+++ b/cfep-23.md
@@ -1,7 +1,7 @@
 <table>
 <tr><td> Title </td><td> Migration to the Zulip chat platform </td>
 <tr><td> Status </td><td> Draft </td></tr>
-<tr><td> Author(s) </td><td> Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt;</td></tr>
+<tr><td> Author(s) </td><td> Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt;</td><td> Matthew R. Becker &lt;becker.mr@gmail.com&gt;</td></tr>
 <tr><td> Created </td><td> Sep 20, 2024</td></tr>
 <tr><td> Updated </td><td> Sep 20, 2024</td></tr>
 <tr><td> Discussion </td><td> https://github.com/conda-forge/cfep/pull/54 </td></tr>
--- a/cfep-23.md
+++ b/cfep-23.md
@@ -18,7 +18,7 @@ Initially, conda-forge offered support through the Gitter platform, which allowe
 
 When Gitter was integrated in the Matrix protocol and migrated to the Element client, it gained encryption features. This meant that security sensitive items could be discussed in the regular private rooms. As a result, Keybase was abandoned and its contents migrated to a secure secret management vault.
 
-The integration in the Element ecosystem was seen as exciting and positive: emoji reactions were available for the first time, and threads were shortly implemented too. However, over time the excitement faded away, which prompted the proposal discussed in this CFEP.
+The integration in the Element ecosystem was seen as exciting and positive: emoji reactions were available for the first time, and threads were shortly implemented too. However, over time usability issues hindered effective communication on Element (described below), which prompted the proposal discussed in this CFEP.
 
 ## Motivation
 
@@ -26,7 +26,7 @@ Element/Matrix offers a free to use real-time chat platform with web, desktop an
 
 However, it suffers from a few issues that limit its usability within our community. The main concerns are:
 
-- End-to-end encryption fails sometimes, leaving the users unable to read previous messages, or have their messaged read by others. Synchronization of keys across clients is necessary to fix it, but in some other cases the issue has dissapeared spontaneously.
+- End-to-end encryption fails sometimes, leaving the users unable to read previous messages, or have their messaged read by others. Synchronization of keys across clients is necessary to fix it, but in some other cases the issue has disappeared spontaneously.
 - Some messages are considered unread for months at a time, and there's no way to find them and mark them as read.
 - The thread system offers very limited ways of tracking conversations and making them discoverable. It's very difficult to follow on topics started hours ago, let alone days or weeks ago.
 - Scoped discussions require a lot of discipline from the users, since threads are optional, and the real-time chat is preferred in the UI.
--- a/cfep-23.md
+++ b/cfep-23.md
@@ -4,7 +4,7 @@
 <tr><td> Author(s) </td><td> Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt;</td></tr>
 <tr><td> Created </td><td> Sep 20, 2024</td></tr>
 <tr><td> Updated </td><td> Sep 20, 2024</td></tr>
-<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda-forge/cfep/pull/54 </td></tr>
 <tr><td> Implementation </td><td> NA </td></tr>
 </table>
 
@@ -67,7 +67,7 @@ Since Zulip doesn't have end-to-end encryption, it is not the best place to disc
 
 ## Alternatives
 
-The community also suggested moving to Discord. However that was discarded given its closed-source nature, handling of the messages there exchanged, and the need of a Discord account. A broader scope discussion can be found at [`conda/communications#29](https://github.com/conda/communications/issues/29).
+The community also suggested moving to Discord. However that was discarded given its closed-source nature, handling of the messages there exchanged, and the need of a Discord account. A broader scope discussion can be found at [`conda/communications#29`](https://github.com/conda/communications/issues/29).
 
 ## Copyright
 
--- a/cfep-23.md
+++ b/cfep-23.md
@@ -50,7 +50,7 @@ After accepting this proposal, we will implement the following decisions:
 - The General channel will be the replacement for the `conda-forge/conda-forge.github.io` chat room. This will be reflected in the conda-forge.org documentation, and reminded periodically in the Element room.
 - The conda-forge core team will move its private communications to the Zulip instance, and stop using the Element rooms. A last message will be published in the Element core chat room with the details on how to transition to the Zulip instance. 
 - New channels will be created in the instance as considered necessary (e.g. infrastructure, compilers, etc).
-- The Zulip instance will be treated as a public message board. No private chat or private rooms for folks besides admins or owners will be enabled.
+- The Zulip instance will be treated as a public message board. No private chat or private rooms for users besides admins, owners, or moderators will be enabled.
 - Sensitive topics will NOT be discussed directly in the Zulip instance. Instead they will follow the protocol described below.
 - Some Element rooms related to conda-forge (e.g., rooms where we interact with other communities like bioconda) will remain unchanged as needed. We will however strive to move as much of the conda-forge Element space to Zulip as possible.
 
--- a/cfep-23.md
+++ b/cfep-23.md
@@ -48,10 +48,11 @@ After accepting this proposal, we will implement the following decisions:
 
 - The Zulip instance published at `conda-forge.zulipchat.com` will be open to the public with world-readable permissions.
 - The General channel will be the replacement for the `conda-forge/conda-forge.github.io` chat room. This will be reflected in the conda-forge.org documentation, and reminded periodically in the Element room.
-- The conda-forge core team will move its private communications to the Zulip instance, and stop using the Element rooms. A last message will be published with the details on how to transition to the Zulip instance.
+- The conda-forge core team will move its private communications to the Zulip instance, and stop using the Element rooms. A last message will be published in the Element core chat room with the details on how to transition to the Zulip instance. 
 - New channels will be created in the instance as considered necessary (e.g. infrastructure, compilers, etc).
 - The Zulip instance will be treated as a public message board. No private chat or private rooms for folks besides admins or owners will be enabled.
 - Sensitive topics will NOT be discussed directly in the Zulip instance. Instead they will follow the protocol described below.
+- Some Element rooms related to conda-forge (e.g., rooms where we interact with other communities like bioconda) will remain unchanged as needed. We will however strive to move as much of the conda-forge Element space to Zulip as possible.
 
 Three months after the approval of this CFEP, the core team will consider the migration finished and the public rooms will become private and archived. If the core team estimates that there's still sufficient activity, another three months will be granted, with periodic reminders of the new Zulip instance. After that period, no more extensions will take place.
 
--- a/cfep-23.md
+++ b/cfep-23.md
@@ -1,0 +1,74 @@
+<table>
+<tr><td> Title </td><td> Migration to the Zulip chat platform </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt;</td></tr>
+<tr><td> Created </td><td> Sep 20, 2024</td></tr>
+<tr><td> Updated </td><td> Sep 20, 2024</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+The conda-forge core team is proposing to transition chat platforms from Element/Matrix to Zulip.
+
+## Background
+
+Initially, conda-forge offered support through the Gitter platform, which allowed one chat room
+per repository. `conda-forge/conda-forge.github.io` was chosen for this and many users have received support over the years in the corresponding room. As of September 2024, over 2000 users have joined the room. The conda-forge core team also used private rooms to communicate core issues. However, due to the lack of encryption in Gitter, a Keybase room was created to exchange information about sensitive topics.
+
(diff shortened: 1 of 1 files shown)


cfep-23.md
```diff
@@ -0,0 +1,74 @@
...
+- Users can "follow" threads they find interesting, which has its own dedicated view. This makes it easier to track updates to conversations you have started or participated in.
+- Private rooms are also available, but they don't offer end-to-end encryption. For this matter, a sensitive communications protocol is proposed below.
+
+## Proposal
+
+After accepting this proposal, we will implement the following decisions:
+
+- The Zulip instance published at `conda-forge.zulipchat.com` will be open to the public with world-readable permissions.
+- The General channel will be the replacement for the `conda-forge/conda-forge.github.io` chat room. This will be reflected in the conda-forge.org documentation, and reminded periodically in the Element room.
+- The conda-forge core team will move its private communications to the Zulip instance, and close the relevant Element rooms.
```
beckermr: We need to leave the old core room open with a note. Many core members are not active and so will loose access otherwise.

cfep-23.md
```diff
@@ -0,0 +1,74 @@
...
+- Sensitive topics will NOT be discussed directly in the Zulip instance. Instead they will follow the protocol described below.
+
+Three months after the approval of this CFEP, the core team will consider the migration finished and the public rooms will become private and archived. If the core team estimates that there's still sufficient activity, another three months will be granted, with periodic reminders of the new Zulip instance. After that period, no more extensions will take place.
+
+## Sensitive communications protocol
+
+Since Zulip doesn't have end-to-end encryption, it is not the best place to discuss sensitive matters. Instead, we propose the following protocol.
+
+1. A private channel will be created in the Zulip instance, with exclusive access to the core team.
+2. In that channel, core members can announce that a new sensitive topic needs to be discussed. A small description describing the situation without revealing sensitive details should be shared.
```
beckermr: We should simplify to say that a new sensitive topic will be announced in a topic in the core channel.

cfep-23.md
beckermr: Also there is a shared conda-forge x bioconda element channel that will have to be kept.

beckermr (dismissed):
I added a few details that need specifying or ironing out.

cfep-23.md
wolfv: An old core member could also reach out via email?

cfep-23.md
```diff
@@ -0,0 +1,74 @@
...
+- Each topic must have a title that promotes well-scoped discussions, which can be marked as solved when the issue has been addressed.
+- Moderation tools are available to split threads into new ones in case the conversation goes off-topic.
+- Users can "follow" threads they find interesting, which has its own dedicated view. This makes it easier to track updates to conversations you have started or participated in.
+- Private rooms are also available, but they don't offer end-to-end encryption. For this matter, a sensitive communications protocol is proposed below.
+
+## Proposal
+
+After accepting this proposal, we will implement the following decisions:
+
+- The Zulip instance published at `conda-forge.zulipchat.com` will be open to the public with world-readable permissions.
```
wolfv, line 48: I think one thing we could think about is wether we want to merge with `conda-community`. I believe having multiple zulip's is a little more friction than multiple element's (because you have to switch subdomains). 

I think I would prefer those two to live close to each other.

cfep-23.md
beckermr: Sure but that core room has been around since conda-forge was created. I'm hesitant to simply delete it.

cfep-23.md
beckermr: I agree with this.

jaimergp: I'm not against this but there are some concerns that would need to be resolved first:

1. Room permissions:
  - The conda community and conda-forge are two separate governed bodies
  - That implies that they will have their separate private rooms to discuss their steering council / core matters
  - Admins will have access to all rooms
  - Hence only members of both governing bodies would be able to have admin permissions. The rest of the teams would need to be moderators and access would need to be restricted. This need to be encoded in the policy.
2. We will need to move the conda community there first, which will require its own CEP, and only then modify this CFEP to vote for our transition to the conda zulip. 
  - Both the CEP and CFEP would have to same the provisions about access control and membership levels. 
  - This would require that we will always need at least one person in both governing bodies. Right now we have (I think) 11 members in common, but who knows how would this look in the future.

Again nothing impossible, but bureaucratically more involved.

ocefpaf: While I'd love to have simpler communication channels for the communities I'm in. I'm afraid this is not a good idea. NumFOCUS review each org independently from time to time and imagine this scenario: One community is no longer a sponsor project but it is still part of a sponsored community communication infrastructure. How would we manage that? Depending on the reasons for the removal of sponsor status, we may be force to "kick out" one of them from zulip.

I don't want to open a can of worms on what-ifs, b/c there are many other fronts where this would be problematic. Also, that scenario is highly unlikely, still, I'd rather not be in that situation.

cfep-23.md
jaimergp: Do you have a link for the conda-forge x bioconda room? I can't find it.

cfep-23.md
beckermr: > Admins will have access to all rooms

This is not true AFAIK. I had to be invited to the steering room for conda even though I am an admin.

cfep-23.md
beckermr: https://matrix.to/#/#bioconda_core-x-conda-forge:gitter.im

cfep-23.md
beckermr: In any case this is not a sticking point for me so happy to go ahead with two spaces.

wolfv: I installed the Zulip Desktop app and it's OK to have two zulips side-by-side that way. Not a blocker from me either, just think the conda community is small enough that it would make sense to have a single spot. But maybe we'll grow!

jaimergp: > This is not true AFAIK. I had to be invited to the steering room for conda even though I am an admin.

@beckermr Can you see a new room named `private-test` from the "Browse / add channels" dialog in the conda Zulip instance? Click on this icon:

image

beckermr: I can see it but I cannot join. 

image

jaimergp: Perfect, thanks for the clarification. I made you an Owner now (apparently it's a different level), does that make a difference @beckermr?

beckermr: No it does not. :/

jakirkham: Jaime think this is a room setting issue. Looking at the description it appears new subscribers can only be added by existing subscribers

Screenshot 2024-09-25 at 1 55 33 PM

<hr>

Separately what is the distinction between Administrators and Owners? Should we just all be made Owners?

jaimergp: Hm, ok, good to know. @beckermr @jakirkham, can you see these buttons in the top part of that dialog that would allow you to delete a channel?

image

jaimergp: > Should we just all be made Owners?

Yes, I don't think there's any reason not to. Right now we are not because... I just learnt about that role distinction as part of this conversation 😬

beckermr: yes I can see the trashcan and it appears that I could click on it, though I did not.

jakirkham: Yes I can see the trashcan, but the subscribe button is still gray

Am unclear whether the latter is fixable. It kind of seems like a Zulip bug atm

Screenshot 2024-09-26 at 1 41 55 PM

<hr>

> > Should we just all be made Owners?
>
> Yes, I don't think there's any reason not to. Right now we are not because... I just learnt about that role distinction as part of this conversation 😬 

All good. We are all learning 😅

jaimergp: It does look while admins and owners cannot see private messages unless explicitly invited by already subscribed users (which is fair), owners can indeed delete the private group chats altogether. Maybe by design, but I don't think it's a bug.

jaimergp: As a result, my concerns of cross-organization visibility are not that severe (owners can't peek!) but they are equally destructive (owners can delete). So if we ever get to a merged conda + conda-forge Zulip community, only members of both entities could be owners.

But since it does look like having two separate instances is not the end of the world, I think we can close this thread here.

cfep-23.md
```diff
@@ -0,0 +1,75 @@
...
+- The General channel will be the replacement for the `conda-forge/conda-forge.github.io` chat room. This will be reflected in the conda-forge.org documentation, and reminded periodically in the Element room.
+- The conda-forge core team will move its private communications to the Zulip instance, and stop using the Element rooms. A last message will be published in the Element core chat room with the details on how to transition to the Zulip instance. 
+- New channels will be created in the instance as considered necessary (e.g. infrastructure, compilers, etc).
+- The Zulip instance will be treated as a public message board. No private chat or private rooms for folks besides admins or owners will be enabled.
```
jaimergp: Shall we include "moderators" here? Thinking of non-core CoC committee members (in the future), staged-recipes, etc.

beckermr: I did not on purpose, but I agree. We can add moderators.

beckermr:
```suggestion
- The Zulip instance will be treated as a public message board. No private chat or private rooms for users besides admins, owners, or moderators will be enabled.
```

beckermr:
@conda-forge/core

This PR falls under the CFEP Vote policy, please vote and/or comment on this PR. This PR needs 60% of core to vote yea to pass.

To vote please leave a PR review according to

 - Approve (yea)
 - Request Changes (nay)
 - Comment w/ "abstain" in it (abstain)

This vote will end on 2024-10-31.

beckermr (approved):

ocefpaf (approved):

jaimergp:
I can't approve my own PR, so here's my vote:

✅  Yes.

mwcraig (approved):

chrisburr (approved):

xhochy (approved):

cfep-23.md
```diff
@@ -0,0 +1,73 @@
...
+- Each topic must have a title that promotes well-scoped discussions, which can be marked as solved when the issue has been addressed.
+- Moderation tools are available to split threads into new ones in case the conversation goes off-topic.
+- Users can "follow" threads they find interesting, which has its own dedicated view. This makes it easier to track updates to conversations you have started or participated in.
+- Private rooms are also available, but they don't offer end-to-end encryption. For this matter, a sensitive communications protocol is proposed below.
+
+## Proposal
+
+After accepting this proposal, we will implement the following decisions:
+
+- The Zulip instance published at `conda-forge.zulipchat.com` will be open to the public with world-readable permissions.
```
viniciusdc, line 48: @jaimergp, you mention the Zulip instance across this doc, but there is no clear definition. My main question when reading the above is whether this `Zulip instance` will work as elementary does right now or if it will be a self-hosted chat service (if it does, where will it reside)?

viniciusdc (approved):

cfep-23.md
jaimergp: Zulipchat.com subdomains are hosted by Zulip, so no, we are not self hosting.

h-vetinari (approved):

marcelotrevisani (approved):

msarahan (approved):

hmaarrfk (approved):

wolfv (approved):

chenghlee (approved):

beckermr:
@conda-forge/core please vote!

jezdez (approved):

ericdill (approved):

minrk (approved):

mariusvniekerk (approved):

beckermr:
@conda-forge/core please vote!

beckermr:
this vote ends tomorrow Oct 31 BTW

loriab (approved):

kkraus14 (approved):

cfep-23.md
```diff
@@ -0,0 +1,73 @@
...
+
+The integration in the Element ecosystem was seen as exciting and positive: emoji reactions were available for the first time, and threads were shortly implemented too. However, over time the excitement faded away, which prompted the proposal discussed in this CFEP.
+
+## Motivation
+
+Element/Matrix offers a free to use real-time chat platform with web, desktop and mobile clients with end-to-end encryption messaging. Each community or space can create one or more rooms, public or private, to discuss different topics. Other features like videocalls, private messaging, and file sharing are available. Users can create a new account or log in with their Github credentials, among others.
+
+However, it suffers from a few issues that limit its usability within our community. The main concerns are:
+
+- End-to-end encryption fails sometimes, leaving the users unable to read previous messages, or have their messaged read by others. Synchronization of keys across clients is necessary to fix it, but in some other cases the issue has dissapeared spontaneously.
```
jakirkham:
```suggestion
- End-to-end encryption fails sometimes, leaving the users unable to read previous messages, or have their messaged read by others. Synchronization of keys across clients is necessary to fix it, but in some other cases the issue has disappeared spontaneously.
```

cfep-23.md
```diff
@@ -0,0 +1,73 @@
...
+- The conda-forge core team will move its private communications to the Zulip instance, and stop using the Element rooms. A last message will be published in the Element core chat room with the details on how to transition to the Zulip instance. 
+- New channels will be created in the instance as considered necessary (e.g. infrastructure, compilers, etc).
+- Sensitive topics will NOT be discussed directly in the Zulip instance. Instead they will follow the protocol described below.
+- Some Element rooms related to conda-forge (e.g., rooms where we interact with other communities like bioconda) will remain unchanged as needed. We will however strive to move as much of the conda-forge Element space to Zulip as possible.
+
+Three months after the approval of this CFEP, the core team will consider the migration finished and the public rooms will become private and archived. If the core team estimates that there's still sufficient activity, another three months will be granted, with periodic reminders of the new Zulip instance. After that period, no more extensions will take place.
```
jakirkham, line 55: Think we would want the history of the public room to remain readable. Would making it private hinder that?

cfep-23.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Sensitive communications protocol
+
+Since Zulip doesn't have end-to-end encryption, it is not the best place to discuss sensitive matters. Instead, we propose the following protocol.
+
+1. Core members will announce that a new sensitive topic needs to be discussed in the Zulip channel for `core`. A small description describing the situation without revealing sensitive details should be shared.
+2. A new private room will be created in Matrix and announced in the private topic. Interested core members can voluntarily join the room to participate or observe the conversation.
+3. When the issue has been addressed, a summary of the events and decisions taken will be captured in the secret vault. The notes document should contain the date, topic title, participants and any other necessary information.
+4. Once the summary has been safely stored, a short announcement should be posted to the Zulip topic informing about the publication of the relevant notes in the vault.
+5. After a short period has passed with no comments, the Element room will be deleted and the Zulip topic will be marked as solved.
```
jakirkham, line 65: Think it is worth noting that the proposed solution for handling sensitive topics relies on Matrix, which we have noted above has trouble validating keys periodically (to be clear Matrix is being overly secure here as opposed to not secure). Ideally we come up with a better solution than relying on Matrix for this

Admittedly there is not a clear alternative atm, but we should investigate moving sensitive communication off Matrix as well. This can be a follow on proposal once we have determined a clear path

cfep-23.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Background
+
+Initially, conda-forge offered support through the Gitter platform, which allowed one chat room per repository. `conda-forge/conda-forge.github.io` was chosen for this and many users have received support over the years in the corresponding room. As of September 2024, over 2000 users have joined the room. The conda-forge core team also used private rooms to communicate core issues. However, due to the lack of encryption in Gitter, a Keybase room was created to exchange information about sensitive topics.
+
+When Gitter was integrated in the Matrix protocol and migrated to the Element client, it gained encryption features. This meant that security sensitive items could be discussed in the regular private rooms. As a result, Keybase was abandoned and its contents migrated to a secure secret management vault.
+
+The integration in the Element ecosystem was seen as exciting and positive: emoji reactions were available for the first time, and threads were shortly implemented too. However, over time the excitement faded away, which prompted the proposal discussed in this CFEP.
```
jakirkham: This reads a little bit like we got bored and so are leaving 😉 This of course is not what is happening

Would suggest we rewrite this like so to call out there were issues. We need not spell them out here as we already do that below

```suggestion
The integration in the Element ecosystem was seen as exciting and positive: emoji reactions were available for the first time, and threads were shortly implemented too. However, over time usability issues hindered effective communication on Element (described below), which prompted the proposal discussed in this CFEP.
```

jakirkham (approved):
Thanks Jaime and everyone who helped evaluate this proposal! 🙏

Approving

Did note a few suggestions below that would be good to include somehow

cfep-23.md
jaimergp: As per Matrix terminology:

image


image


So I'm assuming that private room is the closest we get to read-only (if "Who can read history" is set to "Anyone").

cfep-23.md
jaimergp: I'm hoping that the ephemeral aspect of the room reduces the usability problems. If that proves to be problematic, another CFEP amending this can be introduced to formalize our new preference, I'd say.

beckermr:
This vote has passed. We have 20 yes, 0 no and 0 abstain. It exceeds the 60% quorum requirement of 60% (17 out of 28). Yay!

